Issue #37, opened 2014-10-31
AlignmentRecord not storing "TLEN" / "inferred insert size" from BAM


ryan-williams:
I just noticed this field getting lost when converting into and out of BAMs with ADAM. Should we/I add it? Should we just infer it in ADAM? Or continue not supporting it?

See `TLEN` in the [SAM spec](http://samtools.github.io/hts-specs/SAMv1.pdf).

`htsjdk` puts it in `SAMRecord`s as ["inferred insert size"](https://github.com/samtools/htsjdk/blob/master/src/java/htsjdk/samtools/SAMRecord.java#L538)

fnothaft:
Resolved by #50.

PR #44, opened 2015-02-20
Fragments record type

ilveroluca:
This pull request introduces an Avro record type for Fragments, grouping a number of reads that were sequenced together.

### DIFF
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -57,8 +57,14 @@ record Contig {
   union { null, string } species = null;
 }
 
+record Sequence {
+  union { null, string } bases;
+  union { null, string } qualities;
+}
+
 /**
-   The description of what it has been experimentally read from a DNA fragment.
+   The DNA fragment that is was targeted by the sequencer, resulting in
+   one or more reads.
 */
 record Fragment {
   /**
@@ -79,18 +85,26 @@ record Fragment {
   union { null, int} yPosition = null;
 
   /**
-   The sequence of bases as read by the sequencer.
-   @see qual
+    Fragment's insert size derived from alignment, if the reads have been
+    aligned.
    */
-  array<string> sequences = [];
+  union { null, int} fragmentSize = null;
+
   /**
-   The per-base quality scores of the measured sequences.
+   The sequences read from this fragment.
    */
-  union { null, array<string> } qualities = null;
+  array<Sequence> sequences = [];
+  array<AlignmentRecord> alignments = [];
 }
 
 
 record AlignmentRecord {
+
+  
+  /**
+    Read number within the array of fragment reads.
+  */
+  union { int, null } readNum = 0;
 
   /**
    The reference sequence details for the reference chromosome that
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -67,7 +67,7 @@ record Fragment {
   union { null, string } readName = null;
 
   union { null, string } instrument;
-	union { null, int } runNumber;
+	union { null, string } runId;
   /**
      The following are Illumina specific, is it a problem?
    */
@@ -86,7 +86,7 @@ record Fragment {
   /**
    The per-base quality scores of the measured sequences.
    */
-  union { null, array<string> } qual = null;
+  union { null, array<string> } qualities = null;
 }
   
 
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -30,7 +30,7 @@ record Contig {
    The name of this contig in the assembly (e.g., "chr1").
    */
   union { null, string } contigName = null;
-  
+
   /**
    The length of this contig.
    */
@@ -62,7 +62,7 @@ record Contig {
 */
 record Fragment {
   /**
-     The name of this Fragment. 
+     The name of this Fragment.
   */
   union { null, string } readName = null;
 
@@ -74,10 +74,10 @@ record Fragment {
   union { null, string } flowcellId;
   union { null, int} lane = null;
   union { null, int} tile = null;
-  union { null, int} surface = null;      
+  union { null, int} surface = null;
   union { null, int} xPosition = null;
   union { null, int} yPosition = null;
-  
+
   /**
    The sequence of bases as read by the sequencer.
    @see qual
@@ -88,7 +88,7 @@ record Fragment {
    */
   union { null, array<string> } qualities = null;
 }
-  
+
 
 record AlignmentRecord {
 
@@ -147,7 +147,7 @@ record AlignmentRecord {
    The Compact Ideosyncratic Gapped Alignment Report (CIGAR) string that
    describes the local alignment of this read. Contains {length, operator}
    pairs for all contiguous alignment operations. The operators include:
-   
+
    * M, ALIGNMENT_MATCH: An alignment match indicates that a sequence can be
      aligned to the reference without evidence of an INDEL. Unlike the
      SEQUENCE_MATCH and SEQUENCE_MISMATCH operators, the ALIGNMENT_MATCH
@@ -173,7 +173,7 @@ record AlignmentRecord {
    * =, SEQUENCE_MATCH: This operator indicates that this portion of the aligned
      sequence exactly matches the reference (e.g., all bases are equal to the
      reference bases).
-   * X, SEQUENCE_MISMATCH: This operator indicates that this portion of the 
+   * X, SEQUENCE_MISMATCH: This operator indicates that this portion of the
      aligned sequence is an alignment match to the reference, but a sequence
      mismatch (e.g., the bases are not equal to the reference). This can
      indicate a SNP or a read error.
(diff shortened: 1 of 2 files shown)


src/main/resources/avro/bdg.avdl
```diff
@@ -57,8 +57,55 @@ record Contig {
...
+record Fragment {
+  /**
+     The name of this Fragment.
+  */
+  union { null, string } readName = null;
+
+  union { null, string } instrument;
+	union { null, string } runId;
+  /**
+     The following are Illumina specific, is it a problem?
```
laserson: I'm not crazy about this.  Perhaps we can think of a more general way to support vendor-specific data.

laserson:
What is the use case for this data type?

ilveroluca:
Reads are often analysed in groups (e.g., pairs) including all the reads sequenced from the same DNA fragment.  Example applications are read demultiplexing, alignment, duplicate identification, and even quality control.  This `Fragment` data type would keep all reads from the same DNA fragment together, avoiding having to join on the read id for each of those operations.  Moreover, it would make this data format more useful for MapReduce applications, which with the current bdg formats need two-job strategies for some of those examples I mentioned above.

fnothaft:
Jenkins, add to whitelist.

fnothaft:
@laserson if we move to `Fragment`s instead of `AlignmentRecord`s as our data model for reads, a lot of nice things happen. @ilveroluca and I had a nice discussion about this in the context of building MR based aligners, but I was also realizing other nice benefits. Specifically:

* It would fix this issue: https://github.com/bigdatagenomics/adam/issues/484
* It should improve the performance of duplicate marking (we take a somewhat gnarly performance hit since we need to _re_-pair all read pairs), and we could eliminate `org.bdgenomics.adam.models.SingleReadBucket` and clean up _a lot_ of the language we use to describe Mark Duplicates.
* It doesn't impact BQSR. Most things that we can do with reads in parallel, we can do with fragments in parallel.
* It's advantageous for things like [local reassembly](https://github.com/bigdatagenomics/avocado/pull/127) downstream, where we want all the reads in a pair anyways.
* There are arguably some minor negative implications for INDEL realignment when you've got chimeric reads, but I would argue that those implications are pretty small anyways.

@ilveroluca I think there are a few formatting fixes we need on this PR; I'll send a PR your way (probably tomorrow).

tdanford:
I need to look at the code more carefully, but does this duplicate in any way the ``SingleReadBucket`` and ``ReadBucket`` classes in org.bdgenomics.adam.models?

fnothaft:
@tdanford I would like to make this (`Fragment`) the preferred representation, and then delete `SingleReadBucket` and `ReadBucket` upstream.

tdanford:
That's fine, as long as it's explicit.

src/main/resources/avro/bdg.avdl
```diff
@@ -57,8 +57,55 @@ record Contig {
...
 }
 
+record Sequence {
+  union { null, string } bases;
+  union { null, string } qualities;
+}
+
+/**
+   The DNA fragment that is was targeted by the sequencer, resulting in
+   one or more reads.
```
laserson: I think this description could be improved to show the generality of this object.  Perhaps add a note about how this can refer to a "bucket" of reads that derive from the same originating molecule, or that it can model reads generated from a single library member or cluster (in Illumina terminology).

src/main/resources/avro/bdg.avdl
```diff
@@ -57,8 +57,55 @@ record Contig {
...
+  union { null, string } flowcellId;
+  union { null, int} lane = null;
+  union { null, int} tile = null;
+  union { null, int} surface = null;
+  union { null, int} xPosition = null;
+  union { null, int} yPosition = null;
+
+  /**
+    Fragment's insert size derived from alignment, if the reads have been
+    aligned.
```
laserson: What if you know the fragment size up-front, from the biochemistry?

src/main/resources/avro/bdg.avdl
```diff
@@ -57,8 +57,55 @@ record Contig {
...
+   */
+  union { null, int} fragmentSize = null;
+
+  /**
+   The sequences read from this fragment.
+   */
+  array<Sequence> sequences = [];
+  array<AlignmentRecord> alignments = [];
```
laserson: Is it necessary to have both `array<Sequence>` and `array<AlignmentRecord>`?

src/main/resources/avro/bdg.avdl
```diff
@@ -57,8 +57,55 @@ record Contig {
...
+
+/**
+   The DNA fragment that is was targeted by the sequencer, resulting in
+   one or more reads.
+*/
+record Fragment {
+  /**
+     The name of this Fragment.
+  */
+  union { null, string } readName = null;
```
laserson: How do you envision `readName` to work?

Separately, maybe we need a `barcodeId` or some other type of identifier, for more general cases. I am thinking of cases where you perform library prep that does some kind of subassembly (e.g., [1]).  Or droplet-seq.

[1]: http://www.nature.com/nmeth/journal/v7/n2/full/nmeth.1416.html

laserson:
Overall, I think this will be a useful data type.

src/main/resources/avro/bdg.avdl
ilveroluca: I figure `readName` should be used like the `read id` field in SAM, but without the typical trailing `/1` or `/2`.  Maybe we should change the name to something simple like `id`or `name`.

src/main/resources/avro/bdg.avdl
tdanford: Or sometimes there's a fragment size distribution, right?  How much detail do we want?

fnothaft: I think this is analogous to the `TLEN` field in SAM (see #37).

src/main/resources/avro/bdg.avdl
fnothaft: The sequence array is the literal string of bases per read. The alignment record would describe alignment parameters (e.g., mapQ, CIGAR, etc).

ilveroluca: `sequences` are the original sequences observed by the sequencer.  `alignments` are the alignments of those original sequences, which themselves contain a `sequence` and a `qual` field.

The reason for the duplication is that the sequence in the alignment record is transformed into a canonical format, as in SAM, where the sequence (and the base qualities) is always represented on the forward strand.  Note that this canonical format could be relatively easily reconstructed at runtime from the original sequence plus the alignment information; storing the sequence in the alignment record is thus a matter of convenience and backwards compatibility, though in principle they could easily be dropped from the AlignmentRecord.

src/main/resources/avro/bdg.avdl
ilveroluca: Suggestions?  We could consider throwing things such as `tile`, `xPos` and `yPos` into a map.

The one field I think needs to be kept is `lane`, as it's needed to segment the data in the flowcell.  Though not all sequencing tech has flowcells that are split into "lanes" (only Illumina and SOLiD [apparently](https://genohub.com/ngs-instrument-guide/)), the concept also exists for 454 (they can use different regions of the PTP to sequence different samples).  A good decision might be to change the `lane` field into a more generic `region` and change the type into a `string`, so we can represent any id.

laserson:
I wonder whether `Fragment` as described here is even general enough.  I imagine the primitive should probably be something that could be as general as

```
record Fragment {
  array<AlignmentRecord> alignments
  map<string> metadata
}
```

There are many instances where the primitive data type for an analysis is a set of reads grouped by some property.  Here's a specific example I'm thinking of that I'd like to see supported: I am encapsulating single B cells in water-in-oil emulsions and performing RT-PCR for a single transcript target.  Each droplet incorporates a droplet-specific barcode and each individual PCR primer incorporates a random barcode.  When I get my data back, the reads need to be grouped by droplet barcode (which would correspond to the `Fragment` object) and the analysis itself would depend on the random barcode sequence, which would presumably be extracted from the `AlignmentRecord` object.

fnothaft:
@laserson I see what you're saying, but I'd also like to get this pull request in so that we can start iterating. Would you be OK with us merging it as is, and then making modifications to make `Fragment` more general in a future pull request?

laserson:
Fine by me to merge for now, but this is still not in good enough shape to be as useful as possible

fnothaft:
OK, I'll merge. @laserson can you open an issue or submit a pull request for further discussion?

fnothaft:
Thanks @ilveroluca and @elzaggo! I've manually squashed this down and merged as https://github.com/bigdatagenomics/bdg-formats/commit/b4a8de9f63c9f180c72893b0a213e67ac0717cc4 and https://github.com/bigdatagenomics/bdg-formats/commit/1d73028f9378d8fb7c8317e412d5a59f08658cee.

src/main/resources/avro/bdg.avdl
```diff
@@ -252,8 +257,39 @@ record AlignmentRecord {
...
+
+/**
+   The DNA fragment that is was targeted by the sequencer, resulting in
+   one or more reads.
+*/
+record Fragment {
+  /**
+   The name of this Fragment.
+   */
+  union { null, string } readName = null;
```
ryan-williams: `readName` feels weird to me here. `fragmentName` instead?

PR #50, opened 2015-03-22
[bdg-formats-37] Adds inferred insert size field.

### LINKED ISSUE #37: AlignmentRecord not storing "TLEN" / "inferred insert size" from BAM

fnothaft:
Resolves #37.

### DIFF
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -255,6 +255,10 @@ record AlignmentRecord {
    mate is unaligned, or if the mate does not exist.
    */
   union { null, Contig } mateContig = null;
+  /**
+   The distance between this read and it's mate as inferred from alignment.
+   */
+  union { null, long } inferredInsertSize = null;
 }
 
 record Sequence {


fnothaft:
Rebased.

ilveroluca:
This field has the same meaning as `Fragment.fragmentSize`.  There can be only one :-)

Come to think of it, `inferredInsertSize` seems like a better name than `fragmentSize`.

fnothaft:
@ilveroluca agreed; however, let's get this in (it resolves a silly issue we've got in ADAM) and then make a later change to harmonize the Fragment and AlignmentRecord types. (Essentially, I'd like to do a bit more profiling before committing to Fragments)

fnothaft:
Manually merged as https://github.com/bigdatagenomics/bdg-formats/commit/e24a8f6b2ae39e07e0f313b43887daf7413f7820.

PR #51, opened 2015-04-03
Clean up VariantCallingAnnotations

fnothaft:
Cleans up and adds documentation to the `VariantCallingAnnotation` class. Also makes a bit of cleanup to `Variant` and `Genotype`. Specifically:

* Promote variant quality to the `Variant` object, instead of the `VariantCallingAnnotation` object
* Convert all Phred scaled fields from `int` to `float`
* Pruned a few fields; these were GATK fields that are underdocumented and possibly deprecated (e.g., haplotypeScore). These can still be stored in the tag map.

### DIFF
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -402,6 +402,12 @@ record StructuralVariant {
 
 record Variant {
   /**
+   The Phred scaled error probability of a variant, given the probabilities of
+   the variant in a population.
+   */
+  union { null, int } variantErrorProbability = null;
+
+  /**
    The reference contig that this variant exists on.
    */
   union { null, Contig } contig = null;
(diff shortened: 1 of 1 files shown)


src/main/resources/avro/bdg.avdl
```diff
@@ -58,6 +58,11 @@ record Contig {
 }
 
 record AlignmentRecord {
+ 
+  /**
+   Read number within the array of fragment reads.
+   */
+  union { int, null } readNum = 0;
```
laserson: This should be `null, int` instead of `int, null`.

fnothaft: As written, it is correct, no? The first type needs to be the type of the default value --> e.g., `0` is an `int`.

src/main/resources/avro/bdg.avdl
```diff
@@ -252,8 +257,39 @@ record AlignmentRecord {
...
 /**
- Enumeration for DNA/RNA bases. For codes outside of ACTGU, see the IUPAC 
+   The DNA fragment that is was targeted by the sequencer, resulting in
+   one or more reads.
+*/
+record Fragment {
+  /**
+   The name of this Fragment.
+   */
+  union { null, string } readName = null;
```
laserson: If there are multiple "reads" in a fragment, is `readName` and appropriate name?

src/main/resources/avro/bdg.avdl
```diff
@@ -252,8 +257,39 @@ record AlignmentRecord {
...
+
+  union { null, string } instrument = null;
```
laserson: What data is stored here?

src/main/resources/avro/bdg.avdl
fnothaft: This comment should probably go on #44; I pulled in the commits from there by accident.

src/main/resources/avro/bdg.avdl
```diff
@@ -252,8 +257,39 @@ record AlignmentRecord {
...
+  union { null, string } runId = null; 
+
+  /**
+   Fragment's insert size derived from alignment, if the reads have been
+   aligned.
+   */
+  union { null, int } fragmentSize = null;
```
laserson: This still feels a bit ambiguous to me.  Do you mean "amplicon size"?

fnothaft:
Rebased.

src/main/resources/avro/bdg.avdl
```diff
@@ -472,44 +478,99 @@ enum GenotypeType {
...
+  /**
+   The Wilcoxon rank-sum test statistic of the base quality scores. The base quality
+   scores are separated by whether or not the base supports the reference or the
+   alternate allele.
+   */
+  union { null, float } baseQRankSum = null;
+
+  /**
+   */
+  union { null, float } clippingRankSum = null;
```
arahuja: Is there a doc for this one?

src/main/resources/avro/bdg.avdl
```diff
@@ -472,44 +478,99 @@ enum GenotypeType {
...
+   */
+  union { null, float } rmsMapQ = null;
+  
+  /**
+   The number of reads at this site with mapping quality equal to 0.
+   */
+  union { null, int } mapq0Reads = null;
+
+  /**
+   The Wilcoxon rank-sum test statistic of the mqpping quality scores. The mapping
```
arahuja: s/mqpping/mapping

src/main/resources/avro/bdg.avdl
```diff
@@ -472,44 +478,99 @@ enum GenotypeType {
...
+    The log-odds ratio of being a true vs. false variant under a trained statistical model.
+    This model can be a multivariate Gaussian mixture, support vector machine, etc.
+    */
+  union { null, float } vqslod = null;
+
+  /**
+   If known, the feature that contributed the most to this variant being classified as
+   a false variant.
+   */
+  union { null, string } culprit = null;
```
arahuja: I could see this dropped, pretty vague, does GATK output something like this?

arahuja:
Is there a reason Phred-scaled values should be floats instead of int?

src/main/resources/avro/bdg.avdl
fnothaft: I believe the GATK does. It's useful if you want to "see" why VQSR is filtering variants out. E.g., if you see that VQSR is filtering most variants out on the basis of depth, you can just use a static read depth filter instead of training a statistical model on all your variant calls. We've got WIP VQSR code for avocado and I think we emit this field.

fnothaft:
@arahuja The big reason to use floats is that we store a lot of probabilities that are not close to 1. While we expect the variant and genotype qualities to be close to 1 (and thus OK to use int phred qualities for), the genotype and non-reference likelihoods are generally << 1, so int Phred scores clip to Phred 0.

I would be OK with changing the variant quality and genotype qualities back to integer Phred scores instead of float Phred scores.

src/main/resources/avro/bdg.avdl
fnothaft: I don't think this field is used anywhere (and I don't have good docs) so I'm going to drop it.

fnothaft:
Just pushed an update to address your comments. LMK what you think.

arahuja:
@fnothaft I agree with that, that the int mapping doesn't work for low probabilties.  In these cases it seems worth saving the probabilties themselves then (if we are going to a float, why even bother with Phred scaling?)

fnothaft:
@arahuja Yeah, fair point. I've had a long mental dialog around whether Phred, log probabilities, or probabilities are better. My personal preference is log probabilities, for a few reasons:

* In avocado, we use log probabilities internally. Specifically, I've seen that for very high coverage homozygous sites, the genotype likelihood for the lower likelihood homozygous genotype will underflow if you aren't using log probabilities, which is why we're using log probabilities.
* VCF supports log probabilties for genotype likelihoods (GL, analogous to PL, which is what our current genotypeLikelihood field is)

Thoughts? Would be interesting to hear why you prefer plain floats.

arahuja:
Nope, also support log-probabilities! Just the Phred + float combo was weird to me, but yes, if you like log-probs, that sounds good to me!

fnothaft:
Excellent! I'll cut over to log probabilities shortly for the likelihoods. I'll move the genotype and variant quality fields back to Int.

fnothaft:
Just updated with the Phred<->log change.

fnothaft:
Rebased.

arahuja:
LGTM!

PR #55, opened 2015-04-30
Platform-specific data

ilveroluca:
This pull requests adds support for representing platform-specific data to the schema.

The idea is to add a "type" field (`Fragment.platformType`, specifically) that tells us with what the platform of the `Fragment` was generated, and then store a platform-specific record in the `Fragment.platformInfo` field.

The change includes an enum to list the known platform types and a new record, `IlluminaHTInfo` for the Illumina platform.  The scheme can easily extended to other platforms by simply adding new enum values and defining additional platform-specific records.

### DIFF
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -262,6 +262,19 @@ record Sequence {
   union { null, string } qualities = null;
 }
 
+enum Platform {
+  IlluminaHT /* Illumina high-throughput */
+}
+
+record IlluminaHTInfo {
+  union { null, string } flowcellId;
+  union { null, int } lane;
+  union { null, int } tile;
+  union { null, int } surface;
+  union { null, int } xPosition;
+  union { null, int } yPosition;
+}
+
 /**
    The DNA fragment that is was targeted by the sequencer, resulting in
    one or more reads.
@@ -286,6 +299,13 @@ record Fragment {
    */
   array<Sequence> sequences = [];
   array<AlignmentRecord> alignments = [];
+
+
+  /**
+   Platform-specific information about this Fragment.
+   */
+  union { null, Platform } platformType;
+  union { null, IlluminaHTInfo } platformInfo;
 }
 
 /**


laserson:
I like this approach better than the previous one where we hard-coded the Illumina-specific fields in the `Fragment` schema.  But my preference in general is to avoid platform-specific details in a general concept such as "Fragment".  To play devil's advocate, what are some use cases that require these fields?

ilveroluca:
Devil's advocates are a necessary evil :-)

Some possible use cases for this information, off the top of my head, would be detecting optical duplicates, detecting lane-specific quality problems in a flowcell, demultiplexing reads based on lane and tag.

Moreover, I would argue that we shouldn't be throwing information away unless we can recalculate it or we can't afford to store it.  In this case, we can't recalculate this stuff and, with an efficient columnar storage system, it should be relatively cheap to store.

I'll add that I'm thinking of the `Fragment` record as an entity to use for both processing and archiving data.  I'd be happy to discuss whether this is a sound idea on the mailing list.

src/main/resources/avro/bdg.avdl
```diff
@@ -286,6 +299,13 @@ record Fragment {
    */
   array<Sequence> sequences = [];
   array<AlignmentRecord> alignments = [];
+
+
+  /**
+   Platform-specific information about this Fragment.
+   */
+  union { null, Platform } platformType;
```
laserson: Do Avro schemas need this extra bit of info or can you simply do a runtime type test on the `platformInfo` field?

src/main/resources/avro/bdg.avdl
```diff
@@ -286,6 +299,13 @@ record Fragment {
...
+  union { null, IlluminaHTInfo } platformInfo;
```
laserson: How do you deal with multiple reads that come from different positions in the flow cell?

ilveroluca: You don't.

The `Fragment` is one piece of DNA attached to one position in the flowcell.  The sequencing technology may be able to read the entire thing, generating a single sequence (hypothetically, something like nanopore tech), or it may read multiple parts generating multiple sequences (e.g., Illumina tech generating multiple reads, which are different parts of the same DNA fragment).

With this definition, reads that come from different parts of the flowcell would be sequences read from different `Fragments`.

laserson:
See #56 for further proposal.

heuermh:
Closing as WontFix

PR #63, opened 2015-09-21
add optional referenceIndex field to Contigs

ryan-williams:


fnothaft:
Merged! Thanks @ryan-williams ! I will cut a 0.6.1.

PR #67, opened 2015-10-21
Adding schema for variant annotations

heuermh:
Proposal for variant annotations in VCF as produced by such tools as [SnpEff](http://snpeff.sourceforge.net/) and Ensembl [Variant Effect Predictor (VEP)](http://www.ensembl.org/info/docs/tools/vep/index.html).

VCF annotation ```ANN``` field doc
http://snpeff.sourceforge.net/VCFannotationformat_v1.0.pdf

Annotation branch at GA4GH schemas
https://github.com/ga4gh/schemas/blob/variation_annotation/src/main/resources/avro/alleleAnnotations.avdl

I will add field-level documentation after changes settle down.

### DIFF
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -712,16 +712,6 @@ record Genotype {
 }
 
 /**
- Estimate of putative impact of a variant.
- */
-enum Impact {
-  HIGH,
-  LOW,
-  MODERATE,
-  MODIFIER // no known effect
-}
-
-/**
  Errors, warnings, or informative messages regarding variant annotation accuracy.
  */
 enum VariantAnnotationMessage {
@@ -811,15 +801,8 @@ record TranscriptEffect {
    variant in the context of the feature identified by featureId.  Must be
    Sequence Ontology (SO, see http://www.sequenceontology.org) term names, e.g.
    stop_gained, missense_variant, synonymous_variant, upstream_gene_variant.
-
-   @see featureId
    */
   array<string> effects = [];
-
-  /**
-   Most significant putative impact.
-   */
-  union { null, Impact } impact = null;
 
   /**
    Common gene name (HGNC), e.g. BRCA2.  May be closest gene if annotation
@@ -858,12 +841,15 @@ record TranscriptEffect {
    */
   union { null, int } total = null;
 
-  // (todo:  GA4GH schema includes HGVS.g but the ANN spec does not)
+  /**
+   HGVS.g description of the variant.  See http://www.hgvs.org/mutnomen/recs-DNA.html.
+   */
+  union { null, string } genomicHgvs = null;
 
   /**
    HGVS.c description of the variant.  See http://www.hgvs.org/mutnomen/recs-DNA.html.
    */
-  union { null, string } cdnaHgvs = null;
+  union { null, string } transcriptHgvs = null;
 
   /**
    HGVS.p description of the variant, if coding.  See http://www.hgvs.org/mutnomen/recs-prot.html.
@@ -891,28 +877,17 @@ record TranscriptEffect {
   union { null, int } cdsLength = null;
 
   /**
-   Protein sequence position (one based, including START, but not STOP).
+   Protein sequence position (one based, includes START but not STOP).
    */
   union { null, int } proteinPosition = null;
 
   /**
-   Protein sequence length in amino acids (one based, including START, but not STOP).
+   Protein sequence length in amino acids (one based, includes START but not STOP).
    *
   union { null, int } proteinLength = null;
 
   /**
-   Distance in base pairs to the feature.  For example,
-
-   * Up/Downstream: Distance to first/last codon number
-   * Intergenic: Distance to closest gene
-   * Distance to closest Intron boundary in exon (+/­ up/downstream). If same, use positive
-   * Distance to closest exon boundary in Intron (+/­ up/downstream)
-   * Distance to first base in MOTIF
-   * Distance to first base in miRNA
-   * Distance to exon­intron boundary in splice_site or splice_region
-   * ChipSeq peak: Distance to summit (or peak center)
-   * Histone mark / Histone state: Distance to summit (or peak center)
-
+   Distance in base pairs to the feature.
    */
   union { null, int } distance = null;
 
@@ -936,11 +911,6 @@ record VariantAnnotation {
    Zero or more transcript effects, per alternate allele and transcript (or other feature).
    */
   array<TranscriptEffect> transcriptEffects = [];
-
-  /**
-   Identifiers for variants colocated with the variant for this annotation (todo: this seems oddly placed).
-   */
-  array<string> colocatedVariants = [];
 }
 
 record DatabaseVariantAnnotation {
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -711,12 +711,236 @@ record Genotype {
   union { null, int }     phaseQuality = null;
 }
 
-record VariantEffect {
-  union { null, string} hgvs = null;
-  union { null, string } referenceAminoAcid = null;
-  union { null, string } alternateAminoAcid = null;
-  union {null, string} geneId = null;
-  union {null, string} transcriptId = null;
+/**
+ Estimate of putative impact of a variant.
+ */
+enum Impact {
+  HIGH,
+  LOW,
+  MODERATE,
+  MODIFIER // no known effect
+}
+
+/**
+ Errors, warnings, or informative messages regarding variant annotation accuracy.
+ */
+enum VariantAnnotationMessage {
+
+  /**
+   Chromosome does not exist in reference genome database. Typically indicates
+   a mismatch between the chromosome names in the input file and the chromosome
+   names used in the reference genome.  Message code E1.
+   */
+  ERROR_CHROMOSOME_NOT_FOUND,
+
+  /**
+   The variant's genomic coordinate is greater than chromosome's length.
+   Message code E2.
+   */
+  ERROR_OUT_OF_CHROMOSOME_RANGE,
+
+  /**
+   The 'REF' field in the input VCF file does not match the reference genome.
+   This warning may indicate a conflict between input data and data from
+   reference genome (for instance is the input VCF was aligned to a different
+   reference genome).  Message code W1.
+   */
+  WARNING_REF_DOES_NOT_MATCH_GENOME,
+
+  /**
+   Reference sequence is not available, thus no inference could be performed.
+   Message code W2.
+   */
+  WARNING_SEQUENCE_NOT_AVAILABLE,
+
+  /**
+   A protein coding transcript having a non­multiple of 3 length. It indicates
+   that the reference genome has missing information about this particular
+   transcript.  Message code W3.
+   */
+  WARNING_TRANSCRIPT_INCOMPLETE,
+
+  /**
+   A protein coding transcript has two or more STOP codons in the middle of
+   the coding sequence (CDS). This should not happen and it usually means the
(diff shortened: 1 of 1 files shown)


src/main/resources/avro/bdg.avdl
```diff
@@ -704,12 +704,83 @@ record Genotype {
...
+  union { null, int } proteinPosition = null;
+  union { null, int } proteinLength = null;
+  union { null, int } distance = null;
+  array<Message> messages = [];
+}
+
+/**
+ Annotations of a variant.
+ */
+record VariantAnnotation {
```
fnothaft: Will `VariantAnnotation` be nested under `DatabaseVariantAnnotation` or are they intended to be distinct?

src/main/resources/avro/bdg.avdl
```diff
@@ -704,12 +704,83 @@ record Genotype {
...
+  union { null, MessageCode } code = null;
+  union { null, MessageType } type = null;
+}
+
+/**
+ Annotation of a variant in the context of a feature, typically a transcript.
+ */
+record TranscriptEffect {
+  union { null, string } alternateBases = null;
+  array<string> effects = [];
```
fnothaft: What will go in here? I'm wondering if this might be a good candidate to be made into an enum.

src/main/resources/avro/bdg.avdl
```diff
@@ -704,12 +704,83 @@ record Genotype {
...
+record Message {
+  union { null, MessageCode } code = null;
+  union { null, MessageType } type = null;
+}
+
+/**
+ Annotation of a variant in the context of a feature, typically a transcript.
+ */
+record TranscriptEffect {
+  union { null, string } alternateBases = null;
```
fnothaft: How does this vary from the `alternateAllele` in `Variant`?

heuermh: Same, it is a foreign key of sorts.  I missed that this schema uses ```alternateAllele```, the GA4GH one uses ```alternateBases```.

src/main/resources/avro/bdg.avdl
heuermh: according to the ANN spec, they must be Sequence Ontology term names, so too complicated for an enum.  GA4GH uses ```OntologyTerm```, but populating such on reading VCF would require a term name to URI lookup.  Not sure I see any advantage over using ```string``` and documenting the restriction.

src/main/resources/avro/bdg.avdl
heuermh: Variant annotations come from a tool like SnpEff or VEP whereas database variant annotations look more like what might come from a dbSNP lookup.

src/main/resources/avro/bdg.avdl
```diff
@@ -704,12 +704,83 @@ record Genotype {
...
+  array<string> effects = [];
+  union { null, Impact } impact = null;
+  union { null, string } geneName = null;
+  union { null, string } geneId = null;
+  union { null, string } featureType = null;
+  union { null, string } featureId = null;
+  union { null, string } biotype = null;
```
heuermh: FYI, here is another restricted field without a well-defined domain.  Values come from here http://vega.sanger.ac.uk/info/about/gene_and_transcript_types.html (which appears to include some Sequence Ontology terms) or may not.

src/main/resources/avro/bdg.avdl
fnothaft: OK, cool.

src/main/resources/avro/bdg.avdl
fnothaft: `biotype`?

src/main/resources/avro/bdg.avdl
fnothaft: Ah, gotcha. Could we document this now?

src/main/resources/avro/bdg.avdl
fnothaft: I'd favor cutting this field out.

src/main/resources/avro/bdg.avdl
```diff
@@ -704,12 +704,83 @@ record Genotype {
...
+  INFO_COMPOUND_ANNOTATION, // I2
+  INFO_NON_REFERENCE_ANNOTATION // I3
+}
+
+/**
+ Errors, warnings, or informative messages regarding variant annotation accuracy.
+ */
+record Message {
+  union { null, MessageCode } code = null;
+  union { null, MessageType } type = null;
```
laserson: are `MessageCode` and `MessageType` perfectly correlated or can they be independent?  If they're really the same thing, why have both of them here?

src/main/resources/avro/bdg.avdl
```diff
@@ -704,12 +704,83 @@ record Genotype {
...
+}
+
```
laserson: The names `Message`, `MessageType`, etc. sound super-general to me though they're only intended for use with variant annotations.  Could we specialize the names as well?  On a related note, this file is starting to get sprawling.  Would it make sense to break it into a few pieces?

src/main/resources/avro/bdg.avdl
laserson: +1

src/main/resources/avro/bdg.avdl
laserson: +1

src/main/resources/avro/bdg.avdl
```diff
@@ -704,12 +704,83 @@ record Genotype {
...
+/**
+ Annotation of a variant in the context of a feature, typically a transcript.
+ */
+record TranscriptEffect {
+  union { null, string } alternateBases = null;
+  array<string> effects = [];
+  union { null, Impact } impact = null;
+  union { null, string } geneName = null;
+  union { null, string } geneId = null;
```
laserson: Maybe nix one of `geneName` and `geneId` if these things can get out of sync.

src/main/resources/avro/bdg.avdl
```diff
@@ -704,12 +704,83 @@ record Genotype {
...
+  union { null, string } featureType = null;
+  union { null, string } featureId = null;
+  union { null, string } biotype = null;
+  union { null, int } rank = null;
+  union { null, int } total = null;
+  union { null, string } cdnaHgvs = null;
+  union { null, string } proteinHvgs = null;
```
laserson: Spelling mistake in the `Hvgs`?

src/main/resources/avro/bdg.avdl
laserson: Do we need clearer names for these objects?

src/main/resources/avro/bdg.avdl
fnothaft: We have both `geneName` and `geneId` in [ADAM's Gene class](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/models/Gene.scala). However, `geneName` is an array...

src/main/resources/avro/bdg.avdl
fnothaft: +1

src/main/resources/avro/bdg.avdl
```diff
@@ -704,12 +704,83 @@ record Genotype {
...
+  array<Message> messages = [];
+}
+
+/**
+ Annotations of a variant.
+ */
+record VariantAnnotation {
+  union { null, Variant } variant;
+  array<TranscriptEffect> transcriptEffects = [];
+  array<string> colocatedVariants = [];
```
laserson: What goes here?

src/main/resources/avro/bdg.avdl
fnothaft: Good point. Perhaps `DatabaseVariantAnnotation` could change to `DatabaseVariantStatistics` (?) and `VariantAnnotation` to `VariantEffectAnnotation`?

src/main/resources/avro/bdg.avdl
```diff
@@ -704,12 +704,83 @@ record Genotype {
...
+  union { null, int } proteinLength = null;
+  union { null, int } distance = null;
+  array<Message> messages = [];
+}
+
+/**
+ Annotations of a variant.
+ */
+record VariantAnnotation {
+  union { null, Variant } variant;
```
laserson: So what do we envision a workflow looks like?  A user will store a dataset with the `VariantAnnotation` objects?  Possibly join them to genotypes?

fnothaft: Yeah, probably the same as how we do with `DatabaseVariantAnnotation`s right now.

src/main/resources/avro/bdg.avdl
laserson: Hmm, but `geneId` feels more "definitive" to me in that case as well.

fnothaft: Agree RE: `geneId` being more definitive.

heuermh: Gene name is HUGO gene name, gene id may be any of the billion equivalent identifiers (Ensembl gene id, refseq, accession number, etc.).  These both come from the ANN spec.

src/main/resources/avro/bdg.avdl
heuermh: Will do

src/main/resources/avro/bdg.avdl
heuermh: This needs to be here to identify which alternate allele the annotation is referring to.  Think of it as
```foreach variant { foreach alternate allele { foreach transcript/feature }}}```.

File: src/main/resources/avro/bdg.avdl (hunk 6 above)
Comment (heuermh) [inline src/main/resources/avro/bdg.avdl]: Yeah I think this sucks.  In java I'd have something like
```java
enum Message {
  ERROR_CHROMOSOME_NOT_FOUND(E1, "Chromosome does not exist..."), ...

  private final String code;
  private final String description;
  private Message(String code, String description) { ... }

  public static Message valueOfCode(final String code) {
    // lookup by code
  }
}
```

Don't know how to do something similar in avro.

src/main/resources/avro/bdg.avdl
laserson: lol, the opposite then :)  Sounds like we should have a definitive field and an optional array of other identifiers then.

src/main/resources/avro/bdg.avdl
heuermh: Yep, the annotation may be the effect of the variant on a protein coding transcript, or any of a number of other feature types (lincRNA, siRNA, polymorphic pseudogene, etc.).

src/main/resources/avro/bdg.avdl
heuermh: variants at the same location with different identifiers, see here for example
http://rest.ensembl.org/vep/human/id/COSM476?content-type=application/json

fnothaft:
@heuermh Is this good to go from your side?

heuermh:
I was waiting on clarification offline from the ANN spec folks for a few things, not sure when that will happen, so this is probably ok for now.

heuermh:
Let me take a look at the remaining todos and see if any progress has been made on the ANN spec or GA4GH since.

heuermh:
Updated per recent changes to GA4GH and sent another email to the ANN spec & GA4GH folks for additional clarification.

heuermh:
No word from other folks, will update this after #76 and/or #77 are merged.

fnothaft:
> No word from other folks, will update this after #76 and/or #77 are merged.

Talking about sequencing, what say you:

* We merge #76 
* We rebase #77 on #76, fix the remaining issues, and merge
* We merge this.

CC @erictu

heuermh:
+1 to the plan

fnothaft:
Merged! Thanks @heuermh!

PR #70, opened 2015-12-03
Move record group metadata out from AlignmentRecord.

fnothaft:


src/main/resources/avro/bdg.avdl
```diff
@@ -64,6 +64,28 @@ record Contig {
   union { null, int } referenceIndex = null;
 }
 
+record RecordGroupMetadata {
+  /**
+   Record group identifier.
+   */
+  union { null, string } recordGroupName = null;
```
heuermh: The prefix in all the fields seems redundant.  And this record could just be called RecordGroup, no?

fnothaft: > The prefix in all the fields seems redundant.

Ah yes, it is... I'll drop the prefix.

> And this record could just be called RecordGroup, no?

I think `RecordGroupMetadata` is slightly preferable, since this record just stores the metadata. Depending on how much of a stickler for terminology you are, the "record group" would actually be the full group of reads that are from the same sample/flow cell/lane/etc. Thoughts?

fnothaft:
Just updated to remove the prefix. This wipes out your comment, @heuermh. For convenience, I've replicated the remaining part here:

> @heuermh And this record could just be called RecordGroup, no?

I think RecordGroupMetadata is slightly preferable, since this record just stores the metadata. Depending on how much of a stickler for terminology you are, the "record group" would actually be the full group of reads that are from the same sample/flow cell/lane/etc. Thoughts?

heuermh:
I suppose you are right.  I am generally not a fan of suffixes like Data, Metadata, Info, Class, ClassInfo, Annotation, Record, Result, etc. which are usually meaningless as qualifiers.  I'll have a go at wiping all those out at a later date.  :)

fnothaft:
@heuermh I agree in general, but don't have a great idea as to how to replace the suffix here...

heuermh:
@massie I suppose then this should wait and be merged in sync with your changes to the bigdatagenomics/adam repository?

massie:
+1

fnothaft:
Wouldn't it make sense to get the schema changes in before @massie's changes?

massie:
It's easy to merge this and cut a release when we're ready. Cutting a release now would make things easier for me, but there will be users (likely few, if any though), who might wonder where there metadata fields went. This is finals week for me so I'm probably not going to be able to wrap up the metadata work until next week.

fnothaft:
Merged as https://github.com/bigdatagenomics/bdg-formats/commit/95f4b5b5a5f671fe8fad6e42a6c51224df8cfcad.

PR #72, opened 2016-04-02
Replaced Contig with ContigName in AlignmentRecord

jpdna:
Related to issue #983 in adam repo
Pulls 'Contig' out of 'AlignmentRecord' for efficiency issues, and replace it with 'ContigName' which is used as an index into the 'SequenceDictionary' which contains the `Contig` metadata details

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 
   <groupId>org.bdgenomics.bdg-formats</groupId>
   <artifactId>bdg-formats</artifactId>
-  <version>0.7.1-SNAPSHOT</version>
+  <version>0.7.2-SNAPSHOT</version>
   <packaging>jar</packaging>
   <name>Big Data Genomics: Avro Formats</name>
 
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -11,7 +11,7 @@
 
   <groupId>org.bdgenomics.bdg-formats</groupId>
   <artifactId>bdg-formats</artifactId>
-  <version>0.7.1-SNAPSHOT</version>
+  <version>0.7.2-SNAPSHOT</version>
```
ryan-williams: `0.7.1` isn't released yet, is it? Why this version bump?

ryan-williams: ah, I see this was to have [https://github.com/bigdatagenomics/adam/pull/988] depend on a unique version it could use.

might be nice to have a test maven server we could upload builds to using their git SHA as their maven version, as is discussed somewhat [here](https://github.com/bigdatagenomics/adam/issues/713). out of scope here though, so nevermind!

fnothaft: +1, that's a good suggestion, and probably wouldn't be too hard to implement.

fnothaft:
LGTM! If no one has an objection, I will merge tomorrow PM.

fnothaft:
Thanks @jpdna! Manually merged as https://github.com/bigdatagenomics/bdg-formats/commit/2229115fa4b81dec7459c5bfa4fb18c059c02d4c.

PR #77, opened 2016-04-21
Flatten Variant into Genotype

erictu:
Flattens entire Variant record into Genotype. Doing so will improve query speed when looking at regions by avoiding nested schemas in Parquet. 

Important things to note:
- This impacts any schemas that previously nested Variant. Right now this is just DatabaseVariantAnnotation, but will impact any other VariantAnnotation schemas, such as those sitting in  #67.

### DIFF
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -604,6 +604,19 @@ record Genotype {
   union { null, Variant } variant = null;
 
   /**
+   The reference contig that this genotype's variant exists on.
+   */
+  union { null, string } contigName = null;
+  /**
+   The 0-based start position of this genotype's variant on the reference contig.
+   */
+  union { null, long } start = null;
+  /**
+   The 0-based, exclusive end position of this genotype's variant on the reference contig.
+   */
+  union { null, long } end = null;
+
+  /**
    Statistics collected at this site, if available.
    */
   union { null, VariantCallingAnnotations } variantCallingAnnotations = null;


src/main/resources/avro/bdg.avdl
```diff
@@ -727,7 +721,7 @@ record VariantEffect {
 }
 
 record DatabaseVariantAnnotation {
-  union { null, Variant } variant;
+  union { null, Genotype } genotype;
```
fnothaft: This shouldn't nest a genotype. We're annotating the effect/statistics of a variant, not the called genotype.

fnothaft:
Thanks for opening this @erictu! A quick question: is the performance degradation associated with the nested fields limited to running a predicate on nested fields? I'm curious as to how query performance varies if we fully flatten the `Genotype` schema (i.e., also flatten out `StructuralVariant` and `VariantCallingAnnotations`) versus just doing this flattening. Do you have numbers for this/could you generate numbers for this?

erictu:
You're right, performance degradation is associated with running a predicate, so really I only need the fields that filter operations are performed on (right now pretty basic, just start, end, alleles, contigName). I'll investigate varying degrees of flattening schemas and see how that works though.

fnothaft:
Thanks @erictu!

erictu:
I made changes so only the `contigName`, `start`, and `end`, are flattened into the Genotype record. That being said, I wasn't sure whether to remove them from the Variant record as well.

fnothaft:
I would push all the `Variant` fields down. I was mostly concerned about pushing the `VariantCallingAnnotation` fields into the schema.

heuermh:
If the problem is performance with predicates over nested fields in Parquet, is it possible to improve things there?  This is a lot to flatten in the name of performance.

fnothaft:
It'd have to be improved in Parquet and in Spark SQL (and in Impala and in etc...). I agree though. I don't love making these changes for perf reasons only.

heuermh:
Would it still help if we only did that described in the comment above (https://github.com/bigdatagenomics/bdg-formats/pull/77#issuecomment-214859424)?

Keep ```contigName```, ```start```, and ```end``` in both schema, that is.

erictu:
Yes, that would be fine.

src/main/resources/avro/bdg.avdl
```diff
@@ -388,75 +388,6 @@ enum StructuralVariantType {
   TANDEM_DUPLICATION
 }
 
-record StructuralVariant {
```
fnothaft: `StructuralVariant` needs to be added back.

src/main/resources/avro/bdg.avdl
```diff
@@ -604,6 +604,19 @@ record Genotype {
   union { null, Variant } variant = null;
 
   /**
+   The reference contig that this genotype's variant exists on.
+   */
+  union { null, Contig } contig = null;
```
heuermh: contig here, or contigName?

fnothaft: This should be `union { null, string } contigName;`.

fnothaft:
@erictu can you rebase and make the fix above to this today? Thanks.

erictu:
should be good!

fnothaft:
Merged! Thanks for the patch and the quick turnaround on the last fix, @erictu!

PR #78, opened 2016-05-05
Increase Avro dependency version to 1.8.0

### LINKED ISSUE #73: Update Avro dependency version to 1.8.0
May contain incompatible API changes.

Release notes
https://issues.apache.org/jira/browse/AVRO-1722?jql=project%20%3D%20AVRO%20AND%20fixVersion%20%3D%201.8.0%20ORDER%20BY%20status%20DESC%2C%20priority%20DESC

Source diff
https://github.com/apache/avro/compare/release-1.7.7...release-1.8.0

heuermh:
Fixes #73

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,18 +64,9 @@
           <artifactId>avro-maven-plugin</artifactId>
           <version>${avro.version}</version>
         </plugin>
-        <plugin>
-          <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-gpg-plugin</artifactId>
-          <version>1.6</version>
-        </plugin>
       </plugins>
     </pluginManagement>
     <plugins>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-gpg-plugin</artifactId>
-      </plugin>
       <plugin>
         <groupId>org.apache.avro</groupId>
         <artifactId>avro-maven-plugin</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.sonatype.oss</groupId>
     <artifactId>oss-parent</artifactId>
-    <version>7</version>
+    <version>9</version>
     <relativePath />
   </parent>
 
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -129,7 +133,7 @@
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-gpg-plugin</artifactId>
-            <version>1.5</version>
+            <version>1.6</version>
```
ryan-williams: worth factoring this out since it's used twice?

heuermh: it needs to be in both since it is overriding the declaration in the release profile from oss-parent.  Or do you just mean use a property?

fnothaft: Do we use the `maven-gpg-plugin` without the profile? If not, let's factor it out of the main section of the POM. If so, I think a property might make sense.

scripts/publish-artifacts.sh
```diff
@@ -11,7 +11,7 @@ sha1=$(git log -1 --pretty=format:%H)
 
 # get avrotools jar
 rm -f avro-tools-${AVRO_VERSION}.jar
-wget http://psg.mtu.edu/pub/apache/avro/avro-${AVRO_VERSION}/java/avro-tools-${AVRO_VERSION}.jar
+wget http://mirror.metrocast.net/apache/avro/avro-${AVRO_VERSION}/java/avro-tools-${AVRO_VERSION}.jar
```
fnothaft: Why'd this change?

pom.xml
ryan-williams: Yea I just meant a property, but what @fnothaft said too!

scripts/publish-artifacts.sh
heuermh: http://psg.mtu.edu doesn't resolve for me

fnothaft: Ah, SGTM then.

pom.xml
heuermh: > Do we use the maven-gpg-plugin without the profile?

I don't believe so.

fnothaft: Let's just chop it in the main path then.

heuermh: Yep, just rebased

fnothaft:
Merged! Thanks @heuermh!

heuermh:
Thanks!

PR #81, opened 2016-05-20
Update maven-javadoc-plugin configuration to not fail build on JDK8

### LINKED ISSUE #80: Release fails under Java 8
Releasing bdg-formats fails under Java 8 because certain Javadoc warnings have changed to errors in Java 8. We should look closer to see if the issues are from the comments we've written inline in our avdl, or if it is caused by the Javadoc generated by Avro. To work around, we can just cut the releases using Java 7.

heuermh:
Fixes #80

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,6 +64,14 @@
           <artifactId>avro-maven-plugin</artifactId>
           <version>${avro.version}</version>
         </plugin>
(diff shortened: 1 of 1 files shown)


fnothaft:
Thanks @heuermh!

PR #82, opened 2016-05-20
Update feature and related records to support GFF3

heuermh:
Maximally disruptive proposal for refactoring feature and related records to support GFF3.

If desirable, I could also propose a minimally disruptive refactoring, but where's the fun in that?

I included a `Contig` &rarr; `contigName` change as elsewhere.

### DIFF
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -974,78 +974,192 @@ record DatabaseVariantAnnotation {
 
 }
 
+/**
+ Strand of an alignment or feature.
+ */
 enum Strand {
-  Forward,
-  Reverse,
-  Independent
+
+  /**
+   Forward ("+") strand.
+   */
+  FORWARD,
+
+  /**
+   Reverse ("-") strand.
+   */
+  REVERSE,
+
+  /**
+   Independent or not stranded (".").
+   */
+  INDEPENDENT,
+
+  /**
+   Strandedness is relevant, but unknown ("?").
+   */
+  UNKNOWN
 }
 
+/**
+ Database cross reference in GFF3 style DBTAG:ID format.
+ */
 record Dbxref {
-  string accession;
-  string db;
+
+  /**
+   Database tag in GFF3 style DBTAG:ID format, e.g. EMBL in EMBL:AA816246.
+   */
+  union { null, string } db;
+
+  /**
+   Accession number in GFF3 style DBTAG:ID format, e.g. AA816246 in EMBL:AA816246.
+   */
+  union { null, string } accession;
 }
 
+/**
+ Ontology term cross reference in GFF3 style DBTAG:ID format.
+ */
+record OntologyTerm {
+
+  /**
+   Ontology abbreviation in GFF3 style DBTAG:ID format, e.g. GO in GO:0046703.
+   */
+  union { null, string } db;
+
+  /**
+   Ontology term accession number or identifer in GFF3 style DBTAG:ID format,
+   e.g. 0046703 in GO:0046703.
+   */
+  union { null, string } accession;
+}
+
+/**
+ Feature, such as those represented in native file formats BED, GFF2/GTF,
+ GFF3, IntervalList, and NarrowPeak.
+ */
 record Feature {
+
   /**
-   Preferably unique ID for this Feature object
+   Identifier for this feature.  ID tag in GFF3.
    */
-  union { null, string } featureId = null;
+  union { null, string } id;
 
   /**
-   The type of feature this is (aka, "track").
-
-   Examples are "conservation", "centipede", "gene"
+   Display name for this feature.  Name tag in GFF3, optional column 4 "name"
+   in BED format.
    */
-  union { null, string } featureType = null;
+  union { null, string } name;
 
   /**
-   The original source for this feature.
-
-   Path/filename/URL and/or the file type (e.g., BED, GFF, etc.)
+   Source of this feature.  Column 2 "source" in GFF3.
    */
-  union { null, string } source = null;
+  union { null, string } source;
 
   /**
-   The contig this feature is located on
+   Feature type.  Column 3 "type" in GFF3.
    */
-  union { null, Contig } contig = null;
+  union { null, string } type;
 
   /**
-   Start coordinate on the contig
+   Contig this feature is located on.  Column 1 "seqid" in GFF3, column 1 "chrom"
+   in BED format.
    */
-  union { null, long } start = null;
+  union { null, string } contigName;
 
   /**
-   End coordinate on the contig
+   Start position for this feature, in 0-based coordinate system with closed-open
+   intervals. This may require conversion from the coordinate system of the native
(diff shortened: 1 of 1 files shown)


fnothaft:
This doesn't seem too disruptive to me. I +1. CC @laserson @tdanford for review.

heuermh:
I needed #81 to build this locally (thanks); may have additional doc changes.

For example, what is going on here in the javadocs?
```
@Deprecated
public String contigName

Deprecated. 
Source of this feature. Column 2 "source" in GFF3. union { null, string } source; /**
Feature type. Column 3 "type" in GFF3. union { null, string } type; /** Contig this
feature is located on. Column 1 "seqid" in GFF3, column 1 "chrom" in BED format.
```

heuermh:
Never mind, fixed it

heuermh:
Pushed work-in-progress commit here https://github.com/bigdatagenomics/adam/pull/998/commits/8c212da69ee941b25ede9896c3250de81c7d6de9

tdanford:
Okay, so two high-level comments: 
* OntologyTerm and Dbxref seem duplicative? Why not just a Dbxref? 
* Feature seems pretty GFF3-specific, is that intended?  Most of those fields aren't present in most feature file formats (and ad hoc formats)

src/main/resources/avro/bdg.avdl
```diff
@@ -974,78 +974,192 @@ record DatabaseVariantAnnotation {
...
-   Preferably unique ID for this Feature object
+   Identifier for this feature.  ID tag in GFF3.
    */
-  union { null, string } featureId = null;
+  union { null, string } id;
 
   /**
-   The type of feature this is (aka, "track").
+   Display name for this feature.  Name tag in GFF3, optional column 4 "name"
+   in BED format.
```
laserson: Add example like in the comments above for `OntologyTerm` etc?

src/main/resources/avro/bdg.avdl
```diff
@@ -974,78 +974,192 @@ record DatabaseVariantAnnotation {
...
+   */
+  union { null, string } name;
 
-   Examples are "conservation", "centipede", "gene"
+  /**
+   Source of this feature.  Column 2 "source" in GFF3.
```
laserson: Example here too

src/main/resources/avro/bdg.avdl
```diff
@@ -974,78 +974,192 @@ record DatabaseVariantAnnotation {
...
    */
-  union { null, string } featureType = null;
+  union { null, string } source;
 
   /**
-   The original source for this feature.
+   Feature type.  Column 3 "type" in GFF3.
```
laserson: And here...etc.

heuermh:
> OntologyTerm and Dbxref seem duplicative? Why not just a Dbxref? 

Per the GFF3 [spec](https://github.com/The-Sequence-Ontology/Specifications/blob/master/gff3.md#ontology-associations-and-db-cross-references):
"Two reserved attributes, Ontology_term and Dbxref, can be used to establish links between a GFF3 feature and a data record contained in another database. Ontology_term is reserved for associations to ontologies, such as the Gene Ontology. Dbxref is used for all other cross references. While there is no firm boundary line between these two concepts, curators tend to treat ontology associations differently and hence ontology terms have been given their own reserved attribute label."

I figure if the data take advantage of the distinction, representing some links as Dbxref and some as Ontology_term, we should not lose that information.

> Feature seems pretty GFF3-specific, is that intended? Most of those fields aren't present in most feature file formats (and ad hoc formats)

Yes, this pull request models GFF3 most closely but also allows for things specific to other formats (i.e. `frame` for GFF2/GTF).  I considered adding `geneId` and `transcriptId` for required `gene_id` and `transcript_id` attribute tags for GFF2/GTF.

The way I understand it, with Parquet nulled fields won't increase storage requirements on disk, so having wider but sparse Feature records shouldn't impact performance.  If overhead in RAM is a concern, one can limit by projection.

laserson:
Overall I'm fine with this patch, though I also am not crazy about tying ourselves too closely to a particular file format.

heuermh:
A minimal version of this patch would update Strand as above, add `frame` as a field so that it isn't lost in GFF2/GTF parsing, and remove the `dbxrefs` field since it is not used consistently across the formats.  That would keep all the interesting bits in the attributes field.

heuermh:
Hold on this one, avro `map<string>` for attributes does not preserve cardinality for repeated keys.  Need to think on it...

tdanford:
Broadly, I concur with @laserson -- I think tying ourselves to a particular file format for the "Feature" record isn't great.  I'd prefer a more stripped down Feature or, failing that, renaming Feature to be more clear what it is and what it represents, either `GFF3Feature` or `MODDbFeature` or something along those lines.  

But we've been over both of these objections before, so if neither of them is acceptable to the broader group then I will accept this PR as is.

heuermh:
I've taken care to document how other formats fit the fields in the schema, so I wouldn't say that it is GFF3-specific.  Rather it was driven by a design principle of preferring nullable fields to attributes in a map.

Copying a comment I made in [another thread](https://github.com/bigdatagenomics/adam/issues/1039#issuecomment-222172499):

From what I understand, storing data in attributes is inefficient given how ADAM and Spark operate on schema records (projections, filters, column compression on disk, SparkSQL queries, etc.). For example, it is not possible to project only a single key from the attributes field.

This inefficiency, plus the note above about `map<string>` for attributes not allowing multiple values for the same key, leads me to believe that design principle is sound.  I am of course open to being proved otherwise.  :)

laserson:
This is the classic tradeoff of how strongly typed we want to be.  It seems to me that one issue is that the read/variant types are more "mature" than the feature type, since I don't think as much software is cranking on the feature types.  Would it be possible to mark it somehow as less mature and more subject to change?

fnothaft:
> Rather it was driven by a design principle of preferring nullable fields to attributes in a map.

I strongly advocate for this design principle. We've promoted many fields in `AlignmentRecord` or `Genotype` according to this philosophy.

> Would it be possible to mark it somehow as less mature and more subject to change?

Sounds reasonable to me; maybe just add a sentence in the docs "The Feature schema is experimental and subject to change"?

heuermh:
In any case, attributes with duplicate keys are going to get clobbered.  Is this a reasonable workaround?
```
  // ...
  array<Attribute> attributes = [];
}

record Attribute {
  string key;
  string value;
}
```

And is there an equivalent to [`Multimap<K,V>`](http://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/Multimap.html) in scala?

heuermh:
In the latest commit to ADAM https://github.com/bigdatagenomics/adam/pull/998/commits/8e2d7864aadde1f199b8007fb48843e618316eba I've added special handling for `geneId`, `transcriptId`, and `exonId` fields and duplicate key support for attributes with reserved keys in the GFF3 specification.

fnothaft:
Thanks @heuermh! Merged.

heuermh:
Thank you!

PR #83, opened 2016-05-27
Add sequence, slice, and read schema

heuermh:
Add sequence, slice, and read schema.

@akmorrow13:
Please let me know if you think these might be easier to work with than `NucleotideContigFragment`.

### DIFF
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -1196,7 +1196,7 @@ record Feature {
 
 /**
  Sample.
-*/
+ */
 record Sample {
 
   /**
@@ -1220,5 +1220,168 @@ record Sample {
    Mixture, and Description from sample meta-information lines in VCF files.
    */
   map<string> attributes = {};
+}
+
+/**
+ Alphabet.
+ */
+enum Alphabet {
+
+  /**
+   DNA alphabet.
+   */
+  DNA,
+
+  /**
+   RNA alphabet.
+   */
+  RNA,
+
+  /**
+   Protein alphabet.
+   */
+  PROTEIN
+}
+
+/**
+ Contiguous sequence from an alphabet, e.g. a DNA contig, an RNA transcript,
+ or a protein translation.
+ */
+record Sequence {
+
+  /**
+   Name of this sequence.
+   */
+  union { null, string } name = null;
+
+  /**
+   Description for this sequence.
+   */
+  union { null, string } description = null;
+
+  /**
+   Alphabet for this sequence, defaults to Alphabet.DNA.
+   */
+  union { Alphabet, null } alphabet = "DNA";
+
+  /**
+   Sequence.
+   */
+  union { null, string } sequence = null;
+
+  /**
+   Length of this sequence.
+   */
+  union { null, long } length = null;
+}
+
+/**
+ View of a contiguous region of a sequence.
+ */
+record Slice { // extends Sequence
+
+  /**
+   Name of the sequence this slice views.
+   */
+  union { null, string } name = null;
+
+  /**
+   Description for the sequence this slice views.
+   */
+  union { null, string } description = null;
+
+  /**
+   Alphabet for the sequence this slice views, defaults to Alphabet.DNA.
+   */
+  union { Alphabet, null } alphabet = "DNA";
+
+  /**
+   Sequence for this slice.
+   */
+  union { null, string } sequence = null;
+
+  /**
+   Start position for this slice on the sequence this slice views, in 0-based coordinate
+   system with closed-open intervals.
+   */
+  union { null, long } start = null;
+
+  /**
+   End position for this slice on the sequence this slice views, in 0-based coordinate
+   system with closed-open intervals.
+   */
+  union { null, long } end = null;
+
+  /**
+   Strand for this slice, if any, defaults to Strand.Independent.
+   */
+  union { Strand, null } strand = "Independent";
+
+  /**
+   Length of this slice.
+   */
+  union { null, long } length = null;
+}
+
+/**
+ Quality score variant.
+ */
+enum QualityScoreVariant {
+
+  /**
+   Sanger and Illumina version &gt;= 1.8 FASTQ quality score variant.
+   */
+  FASTQ_SANGER,
+
+  /**
+   Solexa and Illumina version 1.0 FASTQ quality score variant.
+   */
+  FASTQ_SOLEXA,
+
+  /**
+   Illumina version &gt;= 1.3 and &lt; 1.8 FASTQ quality score variant.
+   */
+  FASTQ_ILLUMINA
+}
+
+/**
+ Sequence with quality scores.
+ */
+record Read { // extends Sequence
+
+  /**
+   Name of this read.
+   */
+  union { null, string } name = null;
+
+  /**
+   Description for this read.
+   */
+  union { null, string } description = null;
+
+  /**
+   Alphabet for this read, defaults to Alphabet.DNA.
+   */
+  union { Alphabet, null } alphabet = "DNA";
+
+  /**
+   Sequence for this read.
+   */
+  union { null, string } sequence = null;
+
+  /**
+   Length of this read.
+   */
+  union { null, long } length = null;
+
+  /**
+   Quality scores for this read.
+   */
+  union { null, string } qualityScores = null;
+
+  /**
+   Quality score variant for this read, defaults to QualityScoreVariant.FASTQ_SANGER.
+   */
+  union { QualityScoreVariant, null } qualityScoreVariant = "FASTQ_SANGER";
 }
 }


fnothaft:
OOC, what's the backstory on these?

On first glance, `Sequence`/`Slice` do indeed seem preferable to `NucleotideContigFragment`.

heuermh:
Something I've been meaning to drop in for a while, possibly helpful in #54, could replace `NucleotideContigFragment` in FASTA support, etc.

A use case in Mango came up in our standup on Thursday, and I volunteered to push something for review.

akmorrow13:
I like this we could use this as a replacement for most cases of AlignmentRecord and nucleotideContigFragment. Does slicing allow you to reconfigure sequence lengths, perhaps to a smaller set of sequences?

heuermh:
Slice is a view on a longer sequence, borrowed from the Ensembl Perl API
http://pre.ensembl.org/info/docs/api/core/core_tutorial.html#slices

In other words, it is the same as ADAM's [ReferenceRegion](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala#L142) but with sequence.

fnothaft:
I see that this is tagged as a prerequisite in https://github.com/bigdatagenomics/adam/issues/1048. CC @akmorrow13 @heuermh, do we need this in ADAM 0.20.0 for mango? If so, then we'll want to land this in the next bdg-formats release.

heuermh:
todo: update this to remove `NucleotideContigFragment`

heuermh:
todo: update this to remove `NucleotideContigFragment`

heuermh:
I see the use case a bit more clearly now:

* Read short sequences in FASTA format, each a full sequence<br/>
&nbsp;&nbsp;
&rarr; `loadSequences(...): SequenceRDD`
* Read short or long sequences in FASTA format, partitioned into slices (say 10kB each)<br/>
&nbsp;&nbsp;
&rarr; `loadSlices(...): SliceRDD`
* Read unaligned sequences with quality scores in FASTQ format<br/>
&nbsp;&nbsp;
&rarr; `loadReads(...): ReadRDD`
* Read aligned sequences with quality scores in SAM/BAM format<br/>
&nbsp;&nbsp;
&rarr; `loadAlignments(...): AlignedRecordRDD`

`SequenceRDD` and `SliceRDD` both extend from/implement `ReferenceFile`, such that they implement a method
```scala
def slice(region: ReferenceRegion): Slice { ... }
```

Note: there might be a case for renaming AlignedRecord to AlignedRead, in which case the method might be `loadAlignedReads(...): AlignedReadRDD`; but of course a SAM/BAM file may contain unaligned reads, and could in fact contain only unaligned reads

akmorrow13:
I think SequenceRDD makes a lot of sense. I am still not clear why we need both SliceRDD and SequenceRDD @heuermh ?

heuermh:
`SliceRDD` handles the use case that `NucleotideContigFragmentRDD` is currently handling, where individual sequences might be too long to partition effectively and need to be chopped up into bite size pieces on load.

Implementing the method `slice(ReferenceRegion)` on `SequenceRDD` would be trivial, whereas it needs to be more clever on `SliceRDD`, recombining adjacent slices if necessary.

akmorrow13:
Do these RDD's already exist or should I start implementing them? I imagine slice for SliceRDD would be similar to the current getReferenceString in NucleotideContigFragmentRDD

heuermh:
I was waiting for review of these schema records first.

I imagine implementing SliceRDD would be mostly search & replace in code paths for NucleotideContigFragment.

SequenceRDD and ReadRDD could be based on FASTA and FASTQ support in Hadoop-BAM or stuff cribbed from elsewhere in ADAM.

I've been in the NucleotideContigFragment code before to implement FASTA output, so if you want to start on this, I should be able to help.

heuermh:
Ping to consider merging this for 0.10 release.  I propose adding these now, with support in ADAM 0.20 or 0.21, and then in bdg-formats 0.11 replacing `NucleotideContigFragment` by `Slice`, with refactoring in ADAM 0.21.

When thinking on a converter API, allowing for non-Apache licensed extensions, it would be nice to have these schema for easier conversion to and from third party libraries such as Biojava.

heuermh:
@fnothaft @akmorrow13 @jpdna Sorry, another ping on this one.  These would be useful for the NMDP/Be The Match collaboration, in that they have a need for modeling protein sequences and short peptide fragments.

fnothaft:
OK, making a review pass now. As long as there's an application for this, I am cool with the new schemas.

src/main/resources/avro/bdg.avdl
```diff
@@ -1221,4 +1221,166 @@ record Sample {
...
+
+  /**
+   RNA alphabet.
+   */
+  RNA,
+
+  /**
+   Protein alphabet.
+   */
+  PROTEIN
```
fnothaft: Would it make sense to have an `OTHER` enum as well? E.g., for DNA + methylation? IDK.

src/main/resources/avro/bdg.avdl
```diff
@@ -1221,4 +1221,166 @@ record Sample {
...
+}
+
+/**
+ Sequence.
```
fnothaft: The docs could be better here ;)

fnothaft: More concretely, this is a record that describes a DNA contig, or an RNA transcript, or a protein amino acid sequence, right? If so, the docs should say this.

src/main/resources/avro/bdg.avdl
```diff
@@ -1221,4 +1221,166 @@ record Sample {
...
+ */
+record Sequence {
+
+  /**
+   Name of this sequence.
+   */
+  union { null, string } name = null;
```
fnothaft: Coming back to our discussion on https://github.com/bigdatagenomics/adam/pull/1198 RE: what is name and what is description, it'd prolly be good to flesh out the docs here. Again, I think that the sequence metadata (database IDs) should fall into the `description` field.

src/main/resources/avro/bdg.avdl
```diff
@@ -1221,4 +1221,166 @@ record Sample {
...
+  union { null, string } sequence = null;
+
+  /**
+   Length of this sequence.
+   */
+  union { null, long } length = null;
+}
+
+/**
+ View on a contiguous region of a sequence.
```
fnothaft: `View on a` reads a bit funny to me.

src/main/resources/avro/bdg.avdl
```diff
@@ -1221,4 +1221,166 @@ record Sample {
...
+ */
+record Slice { // extends Sequence
+
+  /**
+   Name of the sequence this slice views.
```
fnothaft: This should be the same as the name of the `Sequence`, right? In the interest of verbosity, let's doc that.

src/main/resources/avro/bdg.avdl
```diff
@@ -1221,4 +1221,166 @@ record Sample {
...
+   */
+  union { null, string } name = null;
+
+  /**
+   Description for the sequence this slice views.
+   */
+  union { null, string } description = null;
```
fnothaft: How do we expect the `slice` record to be used? If we expect to use it for short slices of sequence (e.g., <500bp), then carrying around metadata is going to be inefficient, and we should probably drop this field.

Actually, we may want to drop this field from `Sequence` as well, and just store the sequence metadata in an associated `Contig` record.

src/main/resources/avro/bdg.avdl
```diff
@@ -1221,4 +1221,166 @@ record Sample {
...
+  union { null, long } start = null;
+
+  /**
+   End position for this slice on the sequence this slice views, in 0-based coordinate
+   system with closed-open intervals.
+   */
+  union { null, long } end = null;
+
+  /**
+   Strand for this slice, if any.  Defaults to Strand.Independent.
```
fnothaft: Nit: prefer single space after periods for consistency with other documentation.

src/main/resources/avro/bdg.avdl
```diff
@@ -1221,4 +1221,166 @@ record Sample {
...
+  /**
+   Illumina version &gt;= 1.3 and &lt; 1.8 FASTQ sequence format variant.
+   */
+  ILLUMINA
+}
+
+/**
+ Sequence with quality scores.
+ */
+record Read { // extends Sequence
```
fnothaft: I'm not really much of a fan of adding this record, as I don't see what it really adds over our `AlignmentRecord` and `Fragment` schemas. There isn't much of an overhead to having the `AlignmentRecord` schema with the additional alignment metadata fields (from prior measurement, those are <1% of the space on disk, read sequence is about 30% of space, quality scores are about 60%, and the remaining 10% is contig name and metadata/map qual --> note that these numbers were before we un-nested `Contig`, so the contig name/metadata portion has gone down), and we'd need to add and maintain import/export/convert paths for another schema. I see the `Fragment` schema as useful because it is a natural view over the data --> we create said view during duplicate marking, and would have use for it in variant calling as well.

If we were to add it, I would like the field names to be consistent with `AlignmentRecord`.

fnothaft:
Minus cleaning up the nits, I would be +1 on merging this without the unaligned `Read` schema.

src/main/resources/avro/bdg.avdl
heuermh: +1

src/main/resources/avro/bdg.avdl
heuermh: `View of a` better?

src/main/resources/avro/bdg.avdl
heuermh: Yeah.  If extension were possible in Avro schema, then `Slice` and `Read` would extend `Sequence`

src/main/resources/avro/bdg.avdl
heuermh: I don't mind keeping the field and setting it to null where appropriate.  The use case for `slice` is the same as `NucleotideContigFragment`, and also for view a region of a sequence with features, say a typical GenBank record, or a `slice` in the Ensembl Perl APIs.

src/main/resources/avro/bdg.avdl
heuermh: There was a world before samtools.  ;)

Storing unaligned reads in SAM or related formats may have some practical benefits, but is conceptually strange for say assembly-specific or metagenomic workflows.  Thus I don't mind the similar representations, at least for a short time.  As with `Slice`, which is mostly equivalent to `NucleotideSequenceFragment`, there isn't really a way to start exploring refactoring in the ADAM codebase until the (experimental) schema make it into formats.

I haven't really looked at the `Fragment` schema and code enough to know how it relates.

heuermh:
What about moving these to a separate .avro file, package `org.bdgenomics.formats.avro.experimental` or something similar?

src/main/resources/avro/bdg.avdl
```diff
@@ -1221,4 +1221,166 @@ record Sample {
...
+
+  /**
+   Description for this sequence.
+   */
+  union { null, string } description = null;
+
+  /**
+   Alphabet for this sequence, defaults to Alphabet.DNA.
+   */
+  union { Alphabet, null } alphabet = "DNA";
```
laserson: Is the order of `Alphabet` and `null` reversed here deliberately?  Is this something unique to enums?

src/main/resources/avro/bdg.avdl
```diff
@@ -1221,4 +1221,166 @@ record Sample {
...
+
+  /**
+   Sequence.
+   */
+  union { null, string } sequence = null;
+
+  /**
+   Length of this sequence.
+   */
+  union { null, long } length = null;
```
laserson: Is this just an optimization to explicitly model the length?  Or is this for cases where you have an unknown sequence with a known length?

src/main/resources/avro/bdg.avdl
```diff
@@ -1221,4 +1221,166 @@ record Sample {
...
+  /**
+   Length of this slice.
+   */
+  union { null, long } length = null;
+}
+
+/**
+ FASTQ sequence format variant.
+ */
+enum FastqVariant {
```
laserson: This seems like the wrong name to me.  If I understand correctly, this isn't really a variant of fastq, but rather a variant of how the quality scores are represented.  Maybe `QualityScoresVariant`?

src/main/resources/avro/bdg.avdl
laserson: (e.g., in #54)

src/main/resources/avro/bdg.avdl
fnothaft: The type of the default value should come first in a union. I.e., if the default value is `null`, then `union { null, Alphabet }` would be correct.

heuermh: The intention is to default to `Alphabet.DNA`

src/main/resources/avro/bdg.avdl
heuermh: Intentionally the same as Biojava enum [FastqVariant](http://biojava.org/docs/api1.9.1/org/biojava/bio/program/fastq/FastqVariant.html) per the OBF FASTQ paper.

src/main/resources/avro/bdg.avdl
```diff
@@ -1221,4 +1221,166 @@ record Sample {
...
+
+/**
+ Alphabet.
+ */
+enum Alphabet {
+
+  /**
+   DNA alphabet.
+   */
+  DNA,
```
heuermh: Currently I'm using it to map [to](https://github.com/heuermh/dishevelled-bio/blob/master/convert/src/main/java/org/dishevelled/bio/convert/biojava/BdgenomicsAlphabetToBiojavaAlphabet.java) and [from](https://github.com/heuermh/dishevelled-bio/blob/master/convert/src/main/java/org/dishevelled/bio/convert/biojava/BiojavaAlphabetToBdgenomicsAlphabet.java) Biojava [Alphabet](http://biojava.org/docs/api1.9.1/org/biojava/bio/symbol/Alphabet.html), which is based on IUPAC, but entirely extensible.  Here I see it as simply informational.  Unless of course we want to fully model sequences as symbol lists where symbols come from alphabets, which was discussed elsewhere.

src/main/resources/avro/bdg.avdl
heuermh: Feel free to pull request a groups-of-reads record to this pull request if you have something in mind.

src/main/resources/avro/bdg.avdl
heuermh: An optimization of sorts, similar to storing `Variant.end` even though it can be calculated from `Variant.start` and `Variant.referenceAllele`.

src/main/resources/avro/bdg.avdl
laserson: Hmm, probably not.  Sounds good as is.

src/main/resources/avro/bdg.avdl
laserson: Oh yeah...

src/main/resources/avro/bdg.avdl
laserson: lol, I guess I'm ambivalent then if we're (partly) trying to be consistent with Biojava etc.

heuermh: Are there any non-FASTQ-format quality scores we might want to use in the `Read` record?

fnothaft:
Merged! Thanks @heuermh!

PR #94, opened 2016-07-26
[BDG-FORMATS-20] Add readme

### LINKED ISSUE #20: We need a proper README for bdg-formats
We should leave some top-level documentation in bdg-formats, as part of the README. 

This should explain, among other things, that 
* the .avdl file is the central set of data structures for ADAM and other downstream projects,
* the actual classes are auto-generated by the Maven build,
* the history of moving this out of adam-formats and into its own downstream formats, and
* links to adam, avocado, and any other projects that might use the data structures defined here.

heuermh:
Fixes #20

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,40 @@
+bdg-formats
+===
+
+### Summary
+
+[Big Data Genomics][BDG] (BDG) bdg-formats provides [Apache Avro][Avro] schema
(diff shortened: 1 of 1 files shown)


fnothaft:
Wonderful! Merged. Thanks @heuermh.

Issue #106, opened 2016-09-17
Improve gVCF support


fnothaft:
Specifically, better schema level support for:
- Symbolic alts
- Quality score ranges
- Info `END` field

heuermh:
Closing as WontFix

PR #108, opened 2016-10-24
Refactor Genotype and GenotypeAnnotation

heuermh:
Starting point for discussion around `Genotype` and `GenotypeAnnotation`.  This is a tear-down-and-rebuild, where `VariantCallingAnnotations` has been removed and the bare minimum put back in.

### DIFF
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -990,29 +990,33 @@ record GenotypeAnnotation {
 record Genotype {
 
   /**
-   The variant for this genotype. May be left unset or set to null for performance reasons.
-   */
-  union { null, Variant } variant = null;
-
-  /**
    The reference contig the variant for this genotype exists on. VCF column 1 "CONTIG".
-   Used when the variant field has been left unset or set to null for performance reasons.
    */
   union { null, string } contigName = null;
 
   /**
    The zero-based start position of the variant for this genotype on the reference contig.
    VCF column 2 "POS" converted to zero-based coordinate system, closed-open intervals.
-   Used when the variant field has been left unset or set to null for performance reasons.
    */
   union { null, long } start = null;
 
   /**
    The zero-based, exclusive end position of the variant for this genotype on the reference
-   contig. Calculated by Variant.start + Variant.referenceAllele.length(). Used when the
-   variant field has been left unset or set to null for performance reasons.
+   contig. Calculated by start + referenceAllele.length().
    */
   union { null, long } end = null;
+
+  /**
+   A string describing the reference allele of the variant for this genotype.
+   VCF column 4 "REF".
+   */
+  union { null, string } referenceAllele = null;
+
+  /**
+   A string describing the alternate allele of the variant for this genotype.
+   VCF column 5 "ALT" split for multi-allelic sites.
+   */
+  union { null, string } alternateAllele = null;
 
   /**
    Sample identifier for this genotype. Join with Sample.sampleId for sample metadata.
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -800,166 +800,71 @@ record Variant {
 }
 
 /**
- An enumeration that describes the allele that corresponds to a genotype.
+ Allele encodings for genotypes.
  */
-enum GenotypeAllele {
+enum Allele {
 
   /**
-   The genotype is the reference allele.
+   Reference allele.
    */
   REF,
 
   /**
-   The genotype is the alternate allele.
+   Alternate allele.
    */
   ALT,
 
   /**
-   The genotype is an unspecified other alternate allele. This occurs in our schema
-   when we have split a multi-allelic genotype into two genotype records.
+   An unspecified other alternate allele. This may occur when the variant for
+   the genotype partially encoded by this allele is split as a multi-allelic site.
    */
   OTHER_ALT,
 
   /**
-   The genotype could not be called.
+   Missing allele.
    */
   NO_CALL
 }
 
 /**
- An enumeration that describes the characteristics of a genotype at a site.
+ Genotype annotation.
  */
-enum GenotypeType {
+record GenotypeAnnotation {
 
   /**
-   All genotypes at this site were called as the reference allele.
+   Conditional genotype quality, encoded as a phred quality −10log10 p(genotype call is wrong
+   conditioned on the site being variant). VCF genotype field reserved key GQ.
    */
-  HOM_REF,
+  union { null, int } genotypeQuality = null;
 
   /**
-   Genotypes at this site were called as multiple different alleles. This
-   most commonly occurs if a diploid sample's genotype contains one reference
-   and one variant allele, but can also occur if the genotype contains multiple
-   alternate alleles.
+   Haplotype quality, encoded as two phred qualities. VCF genotype field reserved key HQ.
    */
-  HET,
+  array<int> haplotypeQuality = [];
 
   /**
-   All genotypes at this site were called as a single alternate allele.
+   Phred-scaled probability that alleles are ordered incorrectly, against all other members in
+   the same phase set. VCF genotype field reserved key PQ.
    */
-  HOM_ALT,
+  union { null, int } phaseSetQuality = null;
 
   /**
-   The genotype could not be called at this site.
+   Root Mean Square (RMS) of the mapping quality of reads for this genotype. VCF genotype field
+   reserved key MQ.
    */
-  NO_CALL
-}
-
-/**
- This record represents all stats that, inside a VCF, are stored outside of the
- sample but are computed based on the samples. For instance, MAPQ0 is an aggregate
- stat computed from all samples and stored inside the INFO line.
- */
-record VariantCallingAnnotations {
+  union { null, int } rmsMappingQuality = null;
 
   /**
-   True if filters were applied for this genotype call. FORMAT field "FT" any value other
-   than the missing value.
+   Per-sample component statistics which comprise the Fisher's Exact Test to detect strand bias.
+   If set, the array should contain the number of reads that support the reference allele on the
+   forward strand, the reference allele on the reverse strand, the alternate allele on the forward
+   strand, and the alternate allele on the reverse strand, in that order. VCF genotype field key SB.
    */
-  union { null, boolean } filtersApplied = null;
+  array<int> strandBias = [];
 
   /**
-   True if all filters for this genotype call passed. FORMAT field "FT" value PASS.
-   */
-  union { null, boolean } filtersPassed = null;
-
-  /**
-   Zero or more filters that failed for this genotype call from FORMAT field "FT".
-   */
-  array<string> filtersFailed = [];
-
-  /**
-   True if the reads covering this site were randomly downsampled to reduce coverage.
-   */
-  union { null, boolean } downsampled = null;
-
-  /**
-   The Wilcoxon rank-sum test statistic of the base quality scores. The base quality
-   scores are separated by whether or not the base supports the reference or the
-   alternate allele.
-   */
-  union { null, float } baseQRankSum = null;
-
-  /**
-   The Fisher's exact test score for the strand bias of the reference and alternate
-   alleles. Stored as a phred scaled probability. Thus, if:
-
-   * a = The number of positive strand reads covering the reference allele
-   * b = The number of positive strand reads covering the alternate allele
-   * c = The number of negative strand reads covering the reference allele
-   * d = The number of negative strand reads covering the alternate allele
-
-   This value takes the score:
-   
-   -10 log((a + b)! * (c + d)! * (a + c)! * (b + d)! / (a! b! c! d! n!)
-
-   Where n = a + b + c + d.
-   */
-  union { null, float } fisherStrandBiasPValue = null;
-
-  /**
-   The root mean square of the mapping qualities of reads covering this site.
-   */
-  union { null, float } rmsMapQ = null;
-  
-  /**
-   The number of reads at this site with mapping quality equal to 0.
-   */
-  union { null, int } mapq0Reads = null;
-
-  /**
-   The Wilcoxon rank-sum test statistic of the mapping quality scores. The mapping
-   quality scores are separated by whether or not the read supported the reference or the
-   alternate allele.
-   */
-  union { null, float } mqRankSum = null;
-
-  /**
-   The Wilcoxon rank-sum test statistic of the position of the base in the read at this site.
-   The positions are separated by whether or not the base supports the reference or the
-   alternate allele.
-   */
-  union { null, float } readPositionRankSum = null;
-
-  /**
-   The log scale prior probabilities of the various genotype states at this site.
-   The number of elements in this array should be equal to the ploidy at this
-   site, plus 1.
-   */
-  array<float> genotypePriors = [];
-
-  /**
-   The log scaled posterior probabilities of the various genotype states at this site,
-   in this sample. The number of elements in this array should be equal to the ploidy at
-   this site, plus 1.
-   */
-  array<float> genotypePosteriors = [];
-
-  /**
-   The log-odds ratio of being a true vs. false variant under a trained statistical model.
-   This model can be a multivariate Gaussian mixture, support vector machine, etc.
-   */
-  union { null, float } vqslod = null;
-
-  /**
-   If known, the feature that contributed the most to this variant being classified as
-   a false variant.
-   */
-  union { null, string } culprit = null;
-
-  /**
-   Additional feature info that doesn't fit into the standard fields above.
-   They are all encoded as (string, string) key-value pairs.
+   Additional genotype attributes that do not fit into the standard fields above.
+   The values are stored as strings, even for flag, integer, and float types.
    */
   map<string> attributes = {};
 }
@@ -970,129 +875,76 @@ record VariantCallingAnnotations {
 record Genotype {
 
   /**
-   The variant called at this site.
+   The variant for this genotype.
    */
   union { null, Variant } variant = null;
 
   /**
-   The reference contig that this genotype's variant exists on.
+   The reference contig the variant for this genotype exists on. VCF column 1 "CONTIG".
+   Used when the variant field has been left unset or set to null for performance reasons.
    */
   union { null, string } contigName = null;
 
   /**
-   The 0-based start position of this genotype's variant on the reference contig.
+   The zero-based start position of the variant for this genotype on the reference contig.
+   VCF column 2 "POS" converted to zero-based coordinate system, closed-open intervals.
+   Used when the variant field has been left unset or set to null for performance reasons.
    */
   union { null, long } start = null;
 
   /**
-   The 0-based, exclusive end position of this genotype's variant on the reference contig.
+   The zero-based, exclusive end position of the variant for this genotype on the reference
+   contig. Calculated by Variant.start + Variant.referenceAllele.length(). Used when the
+   variant field has been left unset or set to null for performance reasons.
    */
   union { null, long } end = null;
 
   /**
-   Statistics collected at this site, if available.
+   Sample identifier for this genotype. Join with Sample.sampleId for sample metadata.
+   VCF header line sample id.
    */
-  union { null, VariantCallingAnnotations } variantCallingAnnotations = null;
+  union { null, string } sampleId = null;
 
   /**
-   The unique identifier for this sample.
+   One or more allele values encoding this genotype. If this genotype is phased, the order of
+   the allele values is significant. VCF genotype field reserved key GT mapped to Allele enum
+   values.
    */
-  union { null, string }  sampleId = null;
+  array<Allele> alleles = [];
 
   /**
-   A description of this sample.
-   */
-  union { null, string }  sampleDescription = null;
-
-  /**
-   A string describing the provenance of this sample and the processing applied
-   in genotyping this sample.
-   */
-  union { null, string }  processingDescription = null;
-
-  /**
-   An array describing the genotype called at this site. The length of this
-   array is equal to the ploidy of the sample at this site. This array may
-   reference OTHER_ALT alleles if this site is multi-allelic in this sample.
-   */
-  array<GenotypeAllele> alleles = [];
-
-  /**
-   The expected dosage of the alternate allele in this sample.
-   */
-  union { null, float } expectedAlleleDosage = null;
-
-  /**
-   The number of reads that show evidence for the reference at this site.
-   */
-  union { null, int } referenceReadDepth = null;
-
-  /**
-   The number of reads that show evidence for this alternate allele at this site.
-   */
-  union { null, int } alternateReadDepth = null;
-
-  /**
-   The total number of reads at this site. May not equal (alternateReadDepth +
-   referenceReadDepth) if this site shows evidence of multiple alternate alleles.
-   Analogous to VCF's DP.
-   */
-  union { null, int } readDepth = null;
-
-  /**
-   The minimum number of reads seen at this site across samples when joint
-   calling variants. Analogous to VCF's MIN_DP.
-   */
-  union { null, int } minReadDepth = null;
-
-  /**
-   The phred-scaled probability that we're correct for this genotype call.
-   Analogous to VCF's GQ.
-   */
-  union { null, int } genotypeQuality = null;
-
-  /**
-   Log scaled likelihoods that we have n copies of this alternate allele.
-   The number of elements in this array should be equal to the ploidy at this
-   site, plus 1. Analogous to VCF's PL.
-   */
-  array<double> genotypeLikelihoods = [];
-
-  /**
-   Log scaled likelihoods that we have n non-reference alleles at this site.
-   The number of elements in this array should be equal to the ploidy at this
-   site, plus 1.
-   */
-  array<double> nonReferenceLikelihoods = [];
-
-  /**
-   Component statistics which comprise the Fisher's Exact Test to detect strand bias.
-   If populated, this element should have length 4.
-   */
-  array<int> strandBiasComponents = [];
-
-  /**
-   We split multi-allelic VCF lines into multiple single-alternate records.
-   This bit is set if that happened for this record.
-   */
-  union { boolean, null } splitFromMultiAllelic = false;
-
-  /**
-   True if this genotype is phased.
+   True if this genotype is phased. If true, the order of alleles is significant. VCF genotype field
+   reserved key GT allele separators: '|' genotype is phased; '/' genotype is unphased.
    */
   union { boolean, null } phased = false;
 
   /**
-   The ID of this phase set, if this genotype is phased. Should only be populated
-   if phased == true; else should be null.
+   Phase set identifier, if any. Phased genotypes on the same contig with the same phaseSetId
+   are in the same phased set. All phased genotypes that do not have a phaseSetId are assumed to
+   belong to the same phased set. VCF genotype field reserved key PS.
    */
   union { null, int } phaseSetId = null;
 
   /**
-   Phred scaled quality score for the phasing of this genotype, if this genotype
-   is phased. Should only be populated if phased == true; else should be null.
+   True if filters were applied for this genotype. VCF genotype field reserved key FT any
+   value other than the missing value.
    */
-  union { null, int } phaseQuality = null;
+  union { null, boolean } filtersApplied = null;
+
+  /**
+   True if all filters for this genotype passed. VCF genotype field reserved key FT value PASS.
+   */
+  union { null, boolean } filtersPassed = null;
+
+  /**
+   Zero or more filters that failed for this genotype. VCF genotype field reserved key FT.
+   */
+  array<string> filtersFailed = [];
+
+  /**
+   Annotation for this genotype, if any.
+   */
+  union { null, GenotypeAnnotation } annotation = null;
 }
 
 /**
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -879,15 +879,6 @@ record GenotypeAnnotation {
    */
   union { null, int } readDepth = null;
 
-  // EC Number=A Type=Integer
-  // Comma separated list of expected alternate allele counts for each alternate allele in the same
-  // order as listed in the ALT field. Typically used in association analyses.
-  /**
-   The expected dosage of the alternate allele in this sample. VCF genotype reserved key EC,
-   Number=A, split for multi-allelic sites into a single float value.
-   */
-  union { null, float } expectedAlleleDosage = null; // or expectedAlleleCount? see Variant.alleleCount
-
   // GL Number=G Type=Float
   /**
    Genotype likelihoods comprised of floating point log10-scaled likelihoods for all possible
@@ -895,10 +886,6 @@ record GenotypeAnnotation {
    reserved key GL, Number=G.
    */
   array<float> likelihoods = [];
-
-  // GLE Number=1 Type=String
-  // Genotype likelihoods of heterogeneous ploidy, used in presence of uncertain copy number.
-  // Reserved genotype field key in VCF v4.2, dropped in VCF v4.3
 
   // GP Number=G Type=Float
   /**
@@ -915,25 +902,12 @@ record GenotypeAnnotation {
    */
   union { null, int } quality = null;
 
-  // HQ Number=2 Type=Integer
-  /**
-   Haplotype quality, encoded as two phred qualities. VCF genotype field reserved key HQ, Number=2.
-   */
-  array<int> haplotypeQuality = [];
-
   // MQ Number=1 Type=Integer
   /**
    Root Mean Square (RMS) of the mapping quality of reads for this genotype. VCF genotype field
    reserved key MQ.
    */
   union { null, int } rmsMappingQuality = null;
-
-  // PL Number=G Type=Integer
-  /**
-   Phred-scaled genotype likelihoods rounded to the closest integer, and otherwise defined precisely
-   as the genotype likelihoods field. VCF genotype field reserved key PL, Number=G.
-   */
-  array<int> phredScaledLikelihoods = [];
 
   // PQ Number=1 Type=Integer
   /**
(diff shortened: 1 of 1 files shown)


src/main/resources/avro/bdg.avdl
```diff
@@ -988,6 +703,97 @@ record VariantAnnotation {
...
+  /**
+   Genotype for this annotation.
+   */
+  union { null, Genotype } genotype = null;
+
+  /**
+   Additional genotype attributes that do not fit into the standard fields above.
+   The values are stored as strings, even for flag, integer, and float types.
+   */
+  map<string> attributes = {};
```
fnothaft: We don't require the attributes here; if no attributes are specified, we will default to an empty map. If we wanted to require the user to set attributes, this line would read:

```
map<string> attributes;
```

Our policy is to always allow fields to be nullable. For pure fields, we do this through the nullable syntax. For maps/arrays, we do this by setting the default to an empty map/array.

akmorrow13: ah I see. great!

src/main/resources/avro/bdg.avdl
```diff
@@ -988,6 +703,97 @@ record VariantAnnotation {
...
+   the allele values is significant. VCF genotype field reserved key "GT" mapped to Allele enum
+   values.
+   */
+  array<Allele> alleles = [];
+
+  /**
+   True if this genotype is phased.  If true, the order of alleles is significant. VCF genotype field
+   reserved key "GT" allele separators: '|' genotype is phased; '/' genotype is unphased.
+   */
+  union { boolean, null } phased = false;
```
fnothaft: `phased` doesn't really make sense without the `phaseSetId`, and I'd argue that it isn't great to have without `phaseSetQuality`/`phasingQuality`.

src/main/resources/avro/bdg.avdl
```diff
@@ -988,6 +703,97 @@ record VariantAnnotation {
...
+  /**
+   Missing allele.
+   */
+  NO_CALL
+}
+
+/**
+ Genotype.
+ */
+record Genotype {
```
fnothaft: We lost a lot of pretty important `Genotype` level fields here, like genotype quality, read depth, alt/reference depth, etc. What's the rationale for that?

src/main/resources/avro/bdg.avdl
```diff
@@ -988,6 +703,97 @@ record VariantAnnotation {
...
+
+  /**
+   The variant for this genotype.
+   */
+  union { null, Variant } variant = null;
+
+  /**
```
fnothaft: If we're refactoring `Genotype`, it'd be kinda nice to remove the `contigName`/`start`/`end` fields, since those are nested in `variant`. We flattened them a while back due to a performance regression in Parquet when running predicates.

src/main/resources/avro/bdg.avdl
```diff
@@ -988,6 +703,97 @@ record VariantAnnotation {
...
+   True if this genotype is phased.  If true, the order of alleles is significant. VCF genotype field
+   reserved key "GT" allele separators: '|' genotype is phased; '/' genotype is unphased.
+   */
+  union { boolean, null } phased = false;
+}
+
+/**
+ Genotype annotation.
+ */
+record GenotypeAnnotation {
```
fnothaft: I prefer having `GenotypeAnnotation` nest in `Genotype` than the other way around. The reason I like nesting `Variant` in `VariantAnnotation` is because `Variant` nests in `Genotype`, and thus doing a join between `Genotype`s and `VariantAnnotation`s makes sense to me. Historically, we had `VariantCallingAnnotations` as a nested record solely for convenience. I.e., for a "finished" genotype dataset like 1KG, you would probably not populate the `VariantCallingAnnotations`, since they're just used to determine whether a genotype call is correct or not, and the "finished" callset should be "correct".

src/main/resources/avro/bdg.avdl
```diff
@@ -988,6 +703,97 @@ record VariantAnnotation {
 }
 
 /**
+ Allele encodings for genotypes.
+ */
+enum Allele {
```
fnothaft: +1 on rename and capitalization

heuermh:
> I don't love these changes. I agree with cutting out some of the fields, but the new Genotype record is far too austere for my preferences. Let me take a hack at a PR that is more of a gradual refactor, as a comparison point.

As I stated above, this pull request was just a starting point for discussion.  I plan to gradually add back most if not all of the missing fields.

I started with what I thought were the most uncontroversial and non-complicated fields.

src/main/resources/avro/bdg.avdl
heuermh: See comment below

fnothaft:
> As I stated above, this pull request was just a starting point for discussion. I plan to gradually add back most if not all of the missing fields.

Oops! Got it now. Sorry about the confusion. I put a (lazily named) view of what I think would be good at #109. Let me know your thoughts.

src/main/resources/avro/bdg.avdl
heuermh: Are nested fields still a problem for Parquet?

src/main/resources/avro/bdg.avdl
heuermh: This is part of what I don't understand about the current API.

In conference with @akmorrow13 this afternoon I'm starting to lean more towards removing the *Annotation records altogether in favor of merging the fields onto `Variant` and `Genotype` and nulling them out via projection.

I'd like further discussion around this, if only for my own understanding.

src/main/resources/avro/bdg.avdl
heuermh: Note before this was defined as `phaseSetId != null`.  According to the VCF spec, one can have phased alleles without specifying a phase set ID:

"All phased genotypes that do not contain a PS subfield are assumed to belong to the same phased set."

src/main/resources/avro/bdg.avdl
akmorrow13: Yes, nested fields are still a big problem, mostly in the case when we are trying to minimize latency (Mango). One of the motivations for putting in `contigName/start/end` fields was because of this latency issue.

src/main/resources/avro/bdg.avdl
heuermh: I don't see `phaseSetQuality` in the VCF spec.  `phasingQuality` is VCF genotype field reserved key "PQ".  Is that what you meant or did you have something else in mind?

src/main/resources/avro/bdg.avdl
```diff
@@ -955,51 +670,199 @@ record VariantAnnotation {
...
+   reserved key GT allele separators: '|' genotype is phased; '/' genotype is unphased.
+   */
+  union { boolean, null } phased = false;
+
+  /**
+   Phase set identifier, if any. Phased genotypes on the same contig with the same phaseSetId
+   are in the same phased set. All phased genotypes that do not have a phaseSetId are assumed to
+   belong to the same phased set. VCF genotype field reserved key PS.
+   */
+  union { null, string } phaseSetId = null;
```
heuermh: Note that even though the VCF specification types this as Integer, and recommends "a convention of the position of the first variant in the set identifier (although this is not required)", I've seen all sorts of things used in the wild that aren't Integers.  E.g. the GIAB VCF files use [position_ref_alt], `863556_G_A`

ftp://ftp-trace.ncbi.nlm.nih.gov/giab/ftp/release/NA12878_HG001/latest/NA12878_GIAB_highconf_CG-IllFB-IllGATKHC-Ion-Solid-10X_CHROM1-X_v3.3_highconf.vcf.gz

src/main/resources/avro/bdg.avdl
```diff
@@ -955,51 +670,199 @@ record VariantAnnotation {
...
+   */
+  union { null, boolean } filtersApplied = null;
+
+  /**
+   True if all filters for this variant passed. VCF genotype field reserved key FT value PASS.
+   */
+  union { null, boolean } filtersPassed = null;
+
+  /**
+   Zero or more filters that failed for this variant. VCF genotype field reserved key FT.
```
jpdna: should "..that failed for this variant" be "..that failed for this genotype call"?
Also applies to above field.

heuermh: thanks!  fixed it

src/main/resources/avro/bdg.avdl
fnothaft: I think I'd prefer to stick with Int, as that's what the VCF spec states, and what is IMO correct (i.e., if I were to implement this, I would use an Int "UUID"). That being said, I agree that it is pretty common to see the string phase set ID. I don't know what's the correct thing to do here. If we go with the string, then we emit headers that disagree with the VCF spec, which seems wrong. If we go with Int, then we can't handle VCFs that come in from the wild. If we went with say:

```
union { null, int } phaseSetId = null;
union { null, string } phaseSetStringId = null;
```

Then we'd need to figure out at save whether the user provided string/int/no phase set IDs.

I don't think there's a right/wrong answer, but we see this elsewhere (https://github.com/bigdatagenomics/adam/issues/1213), so we should hash out our philosophy.

heuermh: Right, we should go with what @heuermh argued over in #1213.  :)

heuermh:
Things got hairy on the last rebase, so I'm not 100% sure this is currently the way I want it

fnothaft:
SGTM. Ping me when you'd like me to make a review pass.

heuermh:
By this point it is rather hard to follow the diff; I'd suggest looking at the whole `bdg.avro` file and the docs produced by running `mvn javadoc:javadoc`.

There are a few remaining open questions in the comments, including several fields previously from `Genotype` and `VariantCallingAnnotation` that aren't genotype field reserved keys in the VCF specifications.  Please make a case to re-add those that are necessary.

src/main/resources/avro/bdg.avdl
```diff
@@ -748,299 +748,304 @@ record Variant {
 }
 
 /**
- An enumeration that describes the allele that corresponds to a genotype.
+ Allele encodings for genotypes.
  */
-enum GenotypeAllele {
+enum Allele {
```
devin-petersohn: Is it possible (or even realistic) to get rid of enums? Impala doesn't support importing our parquet formats because of them. I doubt this is what we want, but I wanted to open a discussion about it.

fnothaft: Thoughts for an alternative?

devin-petersohn: The suggested fix for Impala was to rewrite using strings instead.

https://issues.apache.org/jira/browse/IMPALA-2525

heuermh: Well, what is Impala?  It looks like there is a workaround (define table columns yourself and set enum columns as STRING) and a patch linked to that issue.

src/main/resources/avro/bdg.avdl
```diff
@@ -780,39 +780,189 @@ enum Allele {
 record GenotypeAnnotation {
 
   /**
+   Per-sample total alternate read depth. VCF genotype reserved key AD, Number=R, split for
+   multi-allelic sites into single integer values for the reference allele
+   (referenceReadDepth) and the alternate allele (alternateReadDepth, this field).
+   */
+  union { null, int } alternateReadDepth = null;
```
heuermh: I will need some help thinking about what these Number=R values on `Genotype`s mean and how they should be split.  I wrote the doc here to match that in `Variant` but am not sure it makes sense.

src/main/resources/avro/bdg.avdl
```diff
@@ -780,39 +780,189 @@ enum Allele {
...
+  /**
+   Genotype likelihoods comprised of floating point log10-scaled likelihoods for all possible
+   genotypes given [the set of alleles defined in the REF and ALT fields]. VCF genotype field
+   reserved key GL, Number=G.
+   */
+  array<float> likelihoods = [];
+
+  // GLE Number=1 Type=String
+  // Genotype likelihoods of heterogeneous ploidy, used in presence of uncertain copy number.
+  // Reserved genotype field key in VCF v4.2, dropped in VCF v4.3
```
heuermh: Should we keep this in even though it was dropped in the draft VCF v4.3 spec?

src/main/resources/avro/bdg.avdl
```diff
@@ -780,39 +780,189 @@ enum Allele {
...
 
+  // SB is not a reserved key in VCF v4.2 or v4.3
   /**
    Per-sample component statistics which comprise the Fisher's Exact Test to detect strand bias.
    If set, the array should contain the number of reads that support the reference allele on the
    forward strand, the reference allele on the reverse strand, the alternate allele on the forward
    strand, and the alternate allele on the reverse strand, in that order. VCF genotype field key SB.
    */
-  array<int> strandBias = [];
+  //array<int> strandBias = [];  Genotype.strandBiasComponents previously
```
heuermh: Please make a case to re-add any of these commented out fields.

src/main/resources/avro/bdg.avdl
fnothaft: I would drop.

src/main/resources/avro/bdg.avdl
fnothaft: I'll take a look. I am -1 on dropping `strandBias`, and I prefer the old name (`strandBiasComponents`). You can't compute the strand bias components without going back to the reads, and the strand bias info is pretty important for variant filtration.

heuermh: Thanks!  There's no hesitation from me to re-add any of these if they are worth keeping around.  I'm +1 to `strandBiasComponents`.

src/main/resources/avro/bdg.avdl
```diff
@@ -748,299 +748,304 @@ record Variant {
...
    */
-  union { null, boolean } downsampled = null;
+  union { null, float } expectedAlleleDosage = null; // or expectedAlleleCount? see Variant.alleleCount
 
+  // GL Number=G Type=Float
   /**
-   The Wilcoxon rank-sum test statistic of the base quality scores. The base quality
-   scores are separated by whether or not the base supports the reference or the
-   alternate allele.
+   Genotype likelihoods comprised of floating point log10-scaled likelihoods for all possible
```
fnothaft: Are we log or log10?

src/main/resources/avro/bdg.avdl
```diff
@@ -748,299 +748,304 @@ record Variant {
...
+  // EC Number=A Type=Integer
+  // Comma separated list of expected alternate allele counts for each alternate allele in the same
+  // order as listed in the ALT field. Typically used in association analyses.
   /**
-   True if the reads covering this site were randomly downsampled to reduce coverage.
+   The expected dosage of the alternate allele in this sample. VCF genotype reserved key EC,
+   Number=A, split for multi-allelic sites into a single float value.
    */
-  union { null, boolean } downsampled = null;
+  union { null, float } expectedAlleleDosage = null; // or expectedAlleleCount? see Variant.alleleCount
```
fnothaft: EC is a field I would drop.

src/main/resources/avro/bdg.avdl
```diff
@@ -748,299 +748,304 @@ record Variant {
...
-   -10 log((a + b)! * (c + d)! * (a + c)! * (b + d)! / (a! b! c! d! n!)
-
-   Where n = a + b + c + d.
-   */
-  union { null, float } fisherStrandBiasPValue = null;
+  // GLE Number=1 Type=String
+  // Genotype likelihoods of heterogeneous ploidy, used in presence of uncertain copy number.
+  // Reserved genotype field key in VCF v4.2, dropped in VCF v4.3
 
+  // GP Number=G Type=Float
```
fnothaft: +1 to omitting.

src/main/resources/avro/bdg.avdl
```diff
@@ -748,299 +748,304 @@ record Variant {
...
   /**
-   The root mean square of the mapping qualities of reads covering this site.
+   Genotype posterior probabilities in the range 0.0 to 1.0 using the same ordering as the
```
fnothaft: IMO all probabilities/likelihoods should be log scaled.

src/main/resources/avro/bdg.avdl
```diff
@@ -748,299 +748,304 @@ record Variant {
...
    */
-  union { null, float } mqRankSum = null;
+  union { null, int } quality = null;
 
+  // HQ Number=2 Type=Integer
   /**
-   The Wilcoxon rank-sum test statistic of the position of the base in the read at this site.
-   The positions are separated by whether or not the base supports the reference or the
-   alternate allele.
+   Haplotype quality, encoded as two phred qualities. VCF genotype field reserved key HQ, Number=2.
```
fnothaft: I am inclined to suggest dropping this field as it is underspecified.

src/main/resources/avro/bdg.avdl
```diff
@@ -748,299 +748,304 @@ record Variant {
...
    */
-  array<float> genotypePriors = [];
+  union { null, int } rmsMappingQuality = null;
 
+  // PL Number=G Type=Integer
   /**
-   The log scaled posterior probabilities of the various genotype states at this site,
-   in this sample. The number of elements in this array should be equal to the ploidy at
-   this site, plus 1.
+   Phred-scaled genotype likelihoods rounded to the closest integer, and otherwise defined precisely
```
fnothaft: GL and PL are equivalent, just different scales. We should only have one, and IMO it should be GL.

src/main/resources/avro/bdg.avdl
```diff
@@ -748,299 +748,304 @@ record Variant {
...
+   as the genotype likelihoods field. VCF genotype field reserved key PL, Number=G.
    */
-  array<float> genotypePosteriors = [];
+  array<int> phredScaledLikelihoods = [];
 
+  // PQ Number=1 Type=Integer
   /**
-   The log-odds ratio of being a true vs. false variant under a trained statistical model.
-   This model can be a multivariate Gaussian mixture, support vector machine, etc.
+   Phred-scaled probability that alleles are ordered incorrectly, against all other members in
```
fnothaft: It should be "ordered correctly", not "ordered incorrectly", right?

src/main/resources/avro/bdg.avdl
```diff
@@ -748,299 +748,304 @@ record Variant {
...
-   Additional feature info that doesn't fit into the standard fields above.
-   They are all encoded as (string, string) key-value pairs.
+   Per-sample component statistics which comprise the Fisher's Exact Test to detect strand bias.
+   If set, the array should contain the number of reads that support the reference allele on the
+   forward strand, the reference allele on the reverse strand, the alternate allele on the forward
+   strand, and the alternate allele on the reverse strand, in that order. VCF genotype field key SB.
    */
-  map<string> attributes = {};
-}
+  //array<int> strandBias = [];  Genotype.strandBiasComponents previously
```
fnothaft: Actually, ignore what I was saying earlier. Since you've added ADF/ADR, I would prefer to use the ADF/ADR fields over using SB.

src/main/resources/avro/bdg.avdl
```diff
@@ -748,299 +748,304 @@ record Variant {
...
+  //
+  // Not present in VCF specification yet present in Genotype
 
-  /**
-   The reference contig that this genotype's variant exists on.
-   */
-  union { null, string } contigName = null;
+  // The minimum number of reads seen at this site across samples when joint calling variants.
+  // Analogous to VCF's MIN_DP.
+  //union { null, int } minReadDepth = null;
```
fnothaft: Necessary for gVCF support, don't remove.

src/main/resources/avro/bdg.avdl
```diff
@@ -748,299 +748,304 @@ record Variant {
...
 
-  /**
-   The 0-based start position of this genotype's variant on the reference contig.
-   */
-  union { null, long } start = null;
+  // Log scaled likelihoods that we have n non-reference alleles at this site. The number of
+  // elements in this array should be equal to the ploidy at this site, plus 1.
+  //array<float> nonReferenceLikelihoods = [];
```
fnothaft: Necessary for gVCF, don't remove.

Also, log or log10?

src/main/resources/avro/bdg.avdl
```diff
@@ -748,299 +748,304 @@ record Variant {
...
 
-  /**
-   The 0-based, exclusive end position of this genotype's variant on the reference contig.
-   */
-  union { null, long } end = null;
+  // We split multi-allelic VCF lines into multiple single-alternate records. This bit is set
+  // if that happened for this record.
+  //union { boolean, null } splitFromMultiAllelic = false;
```
fnothaft: This is removed because it is present in `Variant`, correct? If so, SGTM.

src/main/resources/avro/bdg.avdl
```diff
@@ -748,299 +748,304 @@ record Variant {
...
-  union { null, string }  sampleId = null;
+  //
+  // Not present in VCF specification yet present in VariantCallingAnnotations
 
-  /**
-   A description of this sample.
-   */
-  union { null, string }  sampleDescription = null;
+  // True if the reads covering this site were randomly downsampled to reduce coverage.
+  //union { null, boolean } downsampled = null;
```
fnothaft: Necessary, don't remove.

src/main/resources/avro/bdg.avdl
```diff
@@ -748,299 +748,304 @@ record Variant {
...
 
-  /**
-   An array describing the genotype called at this site. The length of this
-   array is equal to the ploidy of the sample at this site. This array may
-   reference OTHER_ALT alleles if this site is multi-allelic in this sample.
-   */
-  array<GenotypeAllele> alleles = [];
+  // The Fisher's exact test score for the strand bias of the reference and alternate
+  // alleles.
+  //union { null, float } fisherStrandBiasPValue = null;
```
fnothaft: Necessary for genotype filtering, don't remove.

src/main/resources/avro/bdg.avdl
```diff
@@ -748,299 +748,304 @@ record Variant {
...
+ Genotype.
+ */
+record Genotype {
 
   /**
-   The number of reads that show evidence for the reference at this site.
+   The variant for this genotype. May be left unset or set to null for performance reasons.
    */
-  union { null, int } referenceReadDepth = null;
+  union { null, Variant } variant = null;
```
fnothaft: If we're refactoring, we might as well promote the core variant fields here.

src/main/resources/avro/bdg.avdl
heuermh: +1

src/main/resources/avro/bdg.avdl
heuermh: That doc text comes from VCF spec.  The phrase `[the set of alleles defined in the REF and ALT fields]` needs expansion to explain how the values are split for multi-allelic sites.

src/main/resources/avro/bdg.avdl
heuermh: Omit GP, or GLE from lines 847-849 above?  GP is a reserved key in the VCF spec.

src/main/resources/avro/bdg.avdl
fnothaft: Sure, but my question remains. :-p

src/main/resources/avro/bdg.avdl
fnothaft: Ah sorry for the confusion, my comment went on the wrong line. I was +1 to omitting GLE, not GP.

src/main/resources/avro/bdg.avdl
heuermh: This doc text comes from VCF spec.  They do have some clarification here
https://github.com/samtools/hts-specs/blob/master/VCFv4.3.tex#L486
but I don't guarantee the spec uses those definitions consistently, nor do I argue that the representation in VCF is correct.  If we deviate we need to document the differences in representation and the conversion between them.

src/main/resources/avro/bdg.avdl
heuermh: So if we read in VCF with PL specified, we convert and write out VCF with GL specified?  Reasonable, needs to be documented.

src/main/resources/avro/bdg.avdl
heuermh: `minimumReadDepth` ok?

src/main/resources/avro/bdg.avdl
fnothaft: No, I know the text comes from the VCF spec. What I am saying is that _our_ representation for probabilities/likelihood should:

1. Be consistent across all probability/likelihood fields
2. Always be log-scaled (and with a consistent base)

I don't care what the VCF spec says the format for the field is. The VCF spec is terribly inconsistent WRT number formatting. We should not be inconsistent, and should handle what VCF does in the converter.

src/main/resources/avro/bdg.avdl
fnothaft: OK by me

src/main/resources/avro/bdg.avdl
fnothaft: Probably would want to have some config setting that choses between write PL, write GL (and possibly write both).

src/main/resources/avro/bdg.avdl
heuermh: What do you mean?  Get rid of the `variant` reference?  Add more fields from `Variant` to `Genotype`?  There will be collisions.

src/main/resources/avro/bdg.avdl
heuermh: I don't see `splitFromMultiAllelic` in `Variant` or `VariantAnnotation`

src/main/resources/avro/bdg.avdl
heuermh: This doc text comes from VCF spec.

src/main/resources/avro/bdg.avdl
fnothaft: Ah sorry, you're right. What's the rationale for removing this here? If you don't have this field and you have a triallelic site that is `GT=1/1` in a sample, you don't know that there's a second alt at the site from looking at the genotype record.

src/main/resources/avro/bdg.avdl
fnothaft: Sure, but what's our representation?

src/main/resources/avro/bdg.avdl
fnothaft: Flatten the core variant fields into genotype (`referenceAllele`, `alternateAllele`, etc), as we did with `contigName`, `start`, `end`.

src/main/resources/avro/bdg.avdl
heuermh: Yeah if we need it, let's add it back.  It being removed was a consequence of how I approached this pull request (drop everything and re-add as necessary).  Should it be added back to `Variant` or `VariantAnnotation` as well?

src/main/resources/avro/bdg.avdl
heuermh: We pass it on as is https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala#L919

src/main/resources/avro/bdg.avdl
heuermh: There's not much then, the `filter` ones will collide, and all the interesting bits are on `VariantAnnotation`

src/main/resources/avro/bdg.avdl
fnothaft: What's the representation we _want_ to have?

src/main/resources/avro/bdg.avdl
devin-petersohn: I have tried to define the tables myself, and with the complexity involved in our nested objects, etc. it was a nightmare and I ended up putting it off for a while, especially for trying to just achieve compatibility with a single field. If we're expecting people to be able to use Impala with our schemas we need to keep this in mind.

src/main/resources/avro/bdg.avdl
heuermh: As far as I know, likelihoods should be log10.

src/main/resources/avro/bdg.avdl
heuermh: Agreed.  So we're talking about
```
likelihoods
nonReferenceLikelihoods
priorProbabilities  (is this useful?)
posteriorProbabilities
```
all log10-scaled?  Probabilities in VCF spec are "linear-scale probability in the
posterior distribution, which is Pr(Model$|$Data)".  Are you suggesting log10-scaling those too?

src/main/resources/avro/bdg.avdl
heuermh: I don't have an opinion.

src/main/resources/avro/bdg.avdl
fnothaft: Is there any driver to use log10 versus log?

src/main/resources/avro/bdg.avdl
heuermh: As far as I know, likelihoods should be log10.

src/main/resources/avro/bdg.avdl
heuermh: log10 is what I mean when I say "log-scaled", but then I'm a biologist, not a mathematician or statistician  :)

fnothaft: Yeah, I mean, unless there's a significant reason to go `log10`, I would prefer natural log.

jpdna:
Ping - what are the outstanding questions here?

heuermh:
I'll push a commit tomorrow which addresses most of the review comments.  There are still some unresolved issues in the threads above (e.g. log10 vs. log).

src/main/resources/avro/bdg.avdl
heuermh: Let me know if I have this right:

Field | Cardinality | Action 
----- | ---------- | ------  
`likelihoods` | GL Number=G Type=Float | Convert GL from log10 scale to log scale, optionally convert PL from phred scale to log scale
`nonReferenceLikelihoods` | Number="the ploidy at this site, plus 1" Type=Float | Log scaled, create new VCF key
`genotypePriors` | Number=G Type=Float | Log scaled, rename `priors`, create new VCF key
`genotypePosteriors` | Number=G Type=Float | Log scaled, rename `posteriors`, convert GP from log10 scale to log scale
`posteriorProbabilities` | GP Number=G Type=Float | Drop in favor of log scale `posteriors`

src/main/resources/avro/bdg.avdl
```diff
@@ -748,166 +748,187 @@ record Variant {
...
-  union { null, float } readPositionRankSum = null;
+  //
+  // Not present in VCF specification yet present in Genotype
 
   /**
-   The log scale prior probabilities of the various genotype states at this site.
-   The number of elements in this array should be equal to the ploidy at this
-   site, plus 1.
+   Minimum number of reads seen at this site across samples when joint calling variants.
+   VCF genotype field key MIN_DP.
```
heuermh: This definition is not consistent with other definitions I've seen in the wild for `MIN_DP` (or `Min_DP`, or `min_DP`) VCF genotype field keys, e.g.

GATK
`##FORMAT=<ID=MIN_DP,Number=1,Type=Integer,Description="Minimum DP observed within the GVCF block">`

freebayes
`##FORMAT=<ID=MIN_DP,Number=1,Type=Integer,Description=\"Minimum depth in gVCF output block.\">`

src/main/resources/avro/bdg.avdl
fnothaft: This looks correct, with the exception that `nonReferenceLikelihoods` shouldn't go into a new VCF key, but that `nonReferenceLikelihoods` encodes the GL's for the symbolic non-ref (`*`) alt in a gVCF line.

Handling this gets a bit complex, but the way to think of it is that if `alternateAllele` is non-null, we get a VCF line with two alts (`alternateAllele`, and `*`). In the diploid case, we'd get GL's: `likelihoods(0), likelihoods(1), likelihoods(2), nonRefLikelihoods(1), nonRefLikelihoods(2), 0?`. I'm not entirely sure what should go in the last spot, which would be the `GT=1/2` likelihood, which I feel is not well defined. If `alternateAllele` is null, then we get a VCF line with just `*` as an alt and with GL=`nonRefLikelihoods`

src/main/resources/avro/bdg.avdl
jpdna: I agree the comment documentation for this field should include "minimum depth within gVCF block" wording as do the examples @heuermh  lists above.

fnothaft: +1

heuermh: Is it that simple?

"[the minimum]...at this site across all samples" doesn't mean the same thing as "[the minimum]...across a set of non-variant sites grouped together into non-variant block records that represent intervals of sites for which the genotype quality (GQ) is within a certain range or band."  I don't believe we support
```
##GVCFBlock=minGQ=0(inclusive),maxGQ=5(exclusive)
##GVCFBlock=minGQ=20(inclusive),maxGQ=60(exclusive)
##GVCFBlock=minGQ=5(inclusive),maxGQ=20(exclusive)
##GVCFBlock=minGQ=60(inclusive),maxGQ=2147483647(exclusive)
```
header stuff like HaplotypeCaller outputs unless those are passed through unintentionally.

Or am I reading to much into the current description; it meant gVCF blocks all along?

heuermh:
Fixes #106

heuermh:
Most recent changes:
 * likelihoods and probabilities are log-scaled
 * new VCF genotype field keys `priors`, `nonReferenceLikelihoods`, and `fisherStrandBiasPValue`

I have reservations about log-scaling, in that users are not likely to look at the documentation, and will be confused when e.g. GL and `likelihoods` values do not match.

I'm still not sure what to do re: https://github.com/bigdatagenomics/bdg-formats/pull/108#discussion_r112486393 which may be hidden above.  We currently set the `variant` field and the flattened values when reading from VCF, so unless a user projects away the fields they don't want, we're using more RAM than necessary.

heuermh:
Resolved thread around flatting variant fields on `Genotype` by removing `variant` field and adding `alternateAllele` and `referenceAllele`.

heuermh:
Replaced by #176.

Issue #87, opened 2016-07-06
Use separate filtersFailed and filtersPassed arrays for variant quality filters


heuermh:
See https://github.com/bigdatagenomics/adam/issues/194

PR #109, opened 2016-10-27
Refactor Genotype and GenotypeAnnotation

fnothaft:
An alternative proposal for #108. Would be glad to hear people's thoughts. I've broke all my changes out into one commit per logical change, in the hopes that it will make my reasoning a bit clearer.

### DIFF
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -587,17 +587,6 @@ record Genotype {
   union { null, string }  sampleId = null;
 
   /**
-   A description of this sample.
-   */
-  union { null, string }  sampleDescription = null;
-
-  /**
-   A string describing the provenance of this sample and the processing applied
-   in genotyping this sample.
-   */
-  union { null, string }  processingDescription = null;
-
-  /**
    An array describing the genotype called at this site. The length of this
    array is equal to the ploidy of the sample at this site. This array may
    reference OTHER_ALT alleles if this site is multi-allelic in this sample.
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -486,13 +486,6 @@ record GenotypeAnnotation {
   union { null, boolean } downsampled = null;
 
   /**
-   The Wilcoxon rank-sum test statistic of the base quality scores. The base quality
-   scores are separated by whether or not the base supports the reference or the
-   alternate allele.
-   */
-  union { null, float } baseQRankSum = null;
-
-  /**
    The Fisher's exact test score for the strand bias of the reference and alternate
    alleles. Stored as a phred scaled probability. Thus, if:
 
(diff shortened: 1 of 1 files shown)


heuermh:
Regarding the filter fields, elsewhere (#87) we proposed using separate `filtersPassed` and `filtersFailed` arrays, which I believe is better.  It has one problem with regards to VCF though: there is no distinction between variant and genotype filters in the VCF header, only
```
##FILTER=<ID=ID,Description="description">
```
For the FILTER column (mapping to `Variant.filtersPassed` and `Variant.filtersFailed`) and the genotype field reserved key "FT" (mapping to `Genotype.filtersPassed` and `Genotype.filtersFailed`), the value may be PASS, a semicolon-separated list of filter IDs that fail, or the missing value "." if no filters have been applied.

If the value is PASS, to populate `filtersPassed` the VCF header could be interrogated, but there is no way to know if a filter should go in `Variant.filtersPassed` or `Genotype.filtersPassed` or both.

fnothaft:
> Regarding the filter fields, elsewhere (#87) we proposed using separate filtersPassed and filtersFailed arrays, which I believe is better. It has one problem with regards to VCF though: there is no distinction between variant and genotype filters in the VCF header, only

I've gone back and forth on this a bit recently. Having both `filtersPassed` and `filtersFailed` sounds good in concept, but there's a few points against it:

* If one or more filter fails, you don't care that a variant/genotype passed any other filters.
* The filter strings are a lot of additional metadata; since the majority of sites pass, I'd rather not pay the overhead of storing them for passing sites.
* There isn't really a correct way to generate the filtersPassed field from a VCF. Even if you only had one of genotype or variant filters, people frequently apply filters to only SNPs or only INDELs.

Really, all you care about is whether filters were applied and whether they passed. I think we should punt on this one until we get a use case where someone really needs to know the passing filters for a variant/genotype, and then we can solve the numerous problems.

heuermh:
Agreed.

How about, similar to proposal in #194, boolean `filtered` field and String array `filtersFailed` field on both `Variant` and `Genotype`.

fnothaft:
> How about boolean filtersApplied field, boolean filtersPassed field, and String array filtersFailed field on both Variant and Genotype?

+1

src/main/resources/avro/bdg.avdl
```diff
@@ -695,6 +626,17 @@ record Genotype {
    is phased. Should only be populated if phased == true; else should be null.
    */
   union { null, int } phaseQuality = null;
+
+  /**
+   True or false implies that filters were applied and this genotype PASSed or not.
+   While 'null' implies not filters were applied.
```
jpdna: Should we note in the comment here that this value and the next are derived from the FT field of FORMAT?

heuermh: Take a look at #108

fnothaft:
Closing in favor of #108.

PR #110, opened 2016-11-04
Add variant filters for VCF column 7 FILTER

heuermh:
Per comment here https://github.com/bigdatagenomics/bdg-formats/pull/109#issuecomment-258409758

### DIFF
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -404,6 +404,23 @@ record Variant {
   union { null, string } alternateAllele = null;
 
   /**
+   True if filters were applied for this variant. VCF column 7 "FILTER" any value other
+   than the missing value.
+   */
+  union { null, boolean } filtersApplied = null;
+
+  /**
+   True if all filters for this variant passed. VCF column 7 "FILTER" value PASS.
+   */
+  union { null, boolean } filtersPassed = null;
+
+  /**
+   Zero or more filters that failed for this variant. VCF column 7 "FILTER" shared across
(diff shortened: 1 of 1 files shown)


src/main/resources/avro/bdg.avdl
```diff
@@ -404,6 +404,23 @@ record Variant {
   union { null, string } alternateAllele = null;
 
   /**
+   True if filters were applied for this variant. VCF column 7 "FILTER" any value other
+   than the missing value.
+   */
+  union { null, boolean } filtersApplied = null;
```
fnothaft: This SGTM, but is inconsistent with the filter fields in the `Genotype` record (unless we already renamed them, hard to see in the file browser). I think the structure you're using here is slightly better; the `Genotype` record rolls `filtersApplied` and `filtersPassed` into a single field and encodes "filters not applied" by nulling the field. I'd just like the two to be consistent is all.

heuermh: Right, see commit https://github.com/bigdatagenomics/bdg-formats/pull/108/commits/cd69f8d86db26a8f11059a5c32933c96a2eec68e in #108 for the equivalent `Genotype` changes

fnothaft: Ah, OK! So we'll mod the Genotype bits in #108. SGTM!

fnothaft: Is this good to go from your side? If so, I will merge.

fnothaft:
Merged! Thanks @heuermh!

PR #117, opened 2016-11-30
Add forward reference between Variant and VariantAnnotation

### LINKED ISSUE #116: Forward reference between Variant and VariantAnnotation
The reference between `Variant` and `VariantAnnotation` should be forward so that it can be projected away.

heuermh:
Fixes #116 

Moved `somatic` field from `Variant` to `VariantAnnotation` so that VCF INFO header parsing isn't split between both record types.

Prepares for flattening `VariantAnnotation` into `Variant` at a later date.

Pushed as separate commits since the minor change required reordering the record declarations.

### DIFF
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -421,12 +421,9 @@ record Variant {
   array<string> filtersFailed = [];
 
   /**
-   True if this variant call is somatic; in this case, the referenceAllele will
-   have been observed in another sample. VCF INFO reserved key "SOMATIC", Number=0.
-   Until Number=A and Number=R flags are supported by the VCF specification, this value
-   is shared across all alleles in the same VCF record.
+   Annotation for this variant, if any.
    */
-  union { boolean, null } somatic = false;
+  union { null, VariantAnnotation } annotation = null;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


heuermh:
Should squash because the build will fail on the first commit.

+1 on point release, I can't think of anything else that would need to go.

fnothaft:
> Should squash because the build will fail on the first commit.

Cool. That's what I thought.

> +1 on point release, I can't think of anything else that would need to go.

Actually, how about we get the changes ready to go with the snapshot after I merge this, and then we cut the release. That way, we make sure nothing else needs to go.

fnothaft:
Merged! Thank you, @heuermh!

heuermh:
> Actually, how about we get the changes ready to go with the snapshot after I merge this, and then we cut the release. That way, we make sure nothing else needs to go.

There will be a lot of code changes in ADAM due to this change, are you thinking I should make a separate pull request to handle those, or take them on as part of https://github.com/bigdatagenomics/adam/pull/1288?

fnothaft:
> There will be a lot of code changes in ADAM due to this change, are you thinking I should make a separate pull request to handle those, or take them on as part of bigdatagenomics/adam#1288?

I'm fine either way. TBH, I think it might make sense to get all the changes up-to this into bigdatagenomics/adam#1288, cut the 0.21.0 release, and then make these changes and cut a follow on point release immediately. A bit weird of a sequence, but it'd be nice to not dangle the release after a big change. Thoughts? I'm open to whatever.

heuermh:
I'll make a separate pull request tomorrow morning, if only for comparison's sake.  It'll be mostly delete key.

fnothaft:
> I'll make a separate pull request tomorrow morning, if only for comparison's sake. It'll be mostly delete key.

SGTM!

PR #125, opened 2017-02-13
[FORMATS-124] Add field for generic non-reference allele.

### LINKED ISSUE #124: Variant should store whether it has a non-ref model
Was working on a fix for https://github.com/bigdatagenomics/adam/issues/1381, and the long and short of things is that we should add a flag to note that a variant also has a non-ref (gVCF) model attached. This makes the fix to said issue much easier, and also makes this more explicit (right now, we imply it with a null alt allele).

fnothaft:
Resolves #124. Adds a field to cover the case where a variant contains a symbolic non-reference alternate allele.

I'm not sure if I like this approach. Putting it out there to see if other people like it or don't like it.

### DIFF
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -725,6 +725,12 @@ record Variant {
   union { null, string } alternateAllele = null;
 
   /**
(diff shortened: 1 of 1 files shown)


heuermh:
https://github.com/bigdatagenomics/adam/pull/1389 made this unnecessary, closing as WontFix

PR #131, opened 2017-03-27
[FORMATS-115] Remove AlignmentRecord.mateAlignmentEnd

heuermh:
Fixes #115

PR #132, opened 2017-03-27
[FORMATS-127] Use contigName instead of contig

### LINKED ISSUE #127: NucleotideContigFragment uses `contig` instead of `contigName`
https://github.com/bigdatagenomics/bdg-formats/blob/master/src/main/resources/avro/bdg.avdl#L321

heuermh:
Fixes #127

### DIFF
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -26,7 +26,7 @@ protocol BDG {
 record Contig {
 
   /**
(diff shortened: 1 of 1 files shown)


jpdna:
This seems fine to me.

General question - are we able to merge this into `Master` whenever we like or do we need to coordinate with downstream dependencies?

heuermh:
#131 and #132 are ready to go

PR #134, opened 2017-05-12
[FORMATS-114] Add processing description record.

### LINKED ISSUE #114: Add processing description schema
See https://github.com/bigdatagenomics/adam/issues/1257.

fnothaft:
Resolves #114.

### DIFF
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -102,7 +102,7 @@ record ProcessingStep {
 /**
  Record group metadata.
  */
-record RecordGroupMetadata {
+record RecordGroup {
 
   /**
    Record group identifier.
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -64,6 +64,42 @@ record Contig {
 }
 
 /**
+ Description of a computational processing step.
+ */
+record ProcessingStep {
+
+  /**
+   The ID of this processing step.
+   */
+  union { null, string } id = null;
+
+  /**
+   The name of the program used to run this step.
+   */
+  union { null, string } programName = null;
+
+  /**
+   The command line used to run this step.
+   */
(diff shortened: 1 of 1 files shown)


fnothaft:
I'd like to hold on merging this until https://github.com/bigdatagenomics/adam/issues/1520 is in decent shape, but please start reviewing.

src/main/resources/avro/bdg.avdl
```diff
@@ -87,6 +87,47 @@ record RecordGroupMetadata {
   union { null, int } predictedMedianInsertSize = null;
   union { null, string } platform = null;
   union { null, string } platformUnit = null;
+
+  /**
+   The IDs of the processing steps that have been applied to this record group/
```
heuermh: `record group/` &rarr; `record group.`

src/main/resources/avro/bdg.avdl
```diff
@@ -87,6 +87,47 @@ record RecordGroupMetadata {
...
+   */
+  array<string> processingStepIds = [];
```
heuermh: Since this `*..*` association is between `RecordGroupMetadata`, `ProcessingStep` (, and later `Sample`), all of which will not be large in number, why not have a direct reference instead of the id?

`array<ProcessingStep> processingSteps = [];`

src/main/resources/avro/bdg.avdl
heuermh: And as long as we're in here, please consider #137

src/main/resources/avro/bdg.avdl
fnothaft: Yeah, I was on the hedge about this and could've gone either way. Since you prefer it this way, I'll change it to `array<ProcessingStep>`.

fnothaft:
Pushed fixed commits.

heuermh:
Thank you, @fnothaft

PR #139, opened 2017-05-23
[FORMATS-138] Rename NucleotideContigFragment start/end fields.

### LINKED ISSUE #138: Rename start/end fields in NucleotideContigFragment
All other "top level" records use start/end.

fnothaft:
Resolves #138. Makes NucleotideContigFragment start and end fields consistent
with all other record types that map to a single genomic range.

### DIFF
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -369,38 +369,38 @@ record NucleotideContigFragment {
   /**
    The sequence of bases in this fragment.
    */
-  union { null, string } fragmentSequence = null;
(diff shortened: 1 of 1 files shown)


heuermh:
If we go this way, I would also want to change
`fragmentSequence` &rarr; `sequence`
`fragmentNumber` &rarr; `index`
`fragmentLength` &rarr; `length`
`numberOfFragmentsInContig` &rarr; `fragments`

fnothaft:
> If we go this way, I would also want to change

Done!

heuermh:
Thank you, @fnothaft

PR #145, opened 2017-06-17
[FORMATS-144] Move genotype likelihood fields to doubles.

### LINKED ISSUE #144: Floats for genotype likelihoods can underflow
When converting high phred scaled values (e.g., Phred 1000) to log probabilities, we can underflow the precision of a float and should use a double instead.

fnothaft:
Resolves #144.


heuermh:
Thank you, @fnothaft

PR #154, opened 2017-10-11
[FORMATS-43] Rename contig to reference.

### LINKED ISSUE #43: Rename Contig objects
I was writing code that munged `NucleotideContigFragment`s and realized that `Contig` should be changed to `Reference` and `NucleotideContigFragment` should be renamed to `Contig` or `ContigFragment`.

heuermh:
Fixes #43

### DIFF
--- a/src/main/resources/avro/bdg.avdl
+++ b/src/main/resources/avro/bdg.avdl
@@ -20,47 +20,51 @@
 protocol BDG {
 
 /**
- Record for describing a reference assembly. Not used for storing the contents
- of said assembly.
+ A reference is a canonical assembled contig, intended to act as a
+ reference coordinate space for other genomic annotations. A single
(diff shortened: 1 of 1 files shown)


src/main/resources/avro/bdg.avdl
```diff
@@ -20,47 +20,51 @@
...
-   The name of this assembly (e.g., "hg19").
+   All known corresponding accession IDs for this reference in INSDC
+   (GenBank/ENA/DDBJ), which must include a version number, e.g. GCF_000001405.26.
    */
-  union { null, string } assembly = null;
+  array<string> sourceAccessions = [];
 
   /**
-   The species that this assembly is for.
+   The species that this reference is for.
```
heuermh: Note we could use `OntologyTerm` for species, similar to GA4GH `Reference`

PR #167, opened 2018-09-17
[FORMATS-170] Add sampleId to Feature record

akmorrow13:


src/main/resources/avro/bdg.avdl
```diff
@@ -1165,6 +1165,12 @@ record Feature {
    */
   union { null, string } featureId = null;
 
+
+  /**
+   Identifier for this sample.
+   */
+  union { null, string } sampleId = null;
```
akmorrow13: @heuermh do you think this should be an Avro Sample or just a string sampleId?

heuermh: Elsewhere we use `sampleId`, with the expectation that users will join on sampleId against  metadata `samples` &rarr; `Sample.sampleId`.

heuermh: Yup.  I'd suggest updating the doc string to something like
```java
  /**
   Sample identifier for this feature. Join with Sample.sampleId for sample metadata.
   sampleId tag in GFF3.
  */
```
This matches the style used in #108.

heuermh:
Propose to milestone version 0.11.4.

heuermh:
Thank you, @akmorrow13

PR #220, opened 2024-08-13
update avro dependency version to 1.12.0

heuermh:


heuermh:
See
https://issues.apache.org/jira/browse/AVRO-4034

